PR #4, opened 2022-06-21
Add tests for errors handling in tasks

ffevotte:
This checks that:

- [x] errors are not handled when debug mode is off
- [x] errors are handled when debug mode is on
- [x] tasks can be stopped (without error) when debug mode is on

**NB:** This PR also renames `debug_mode` to `_debug_mode` to mark it as part of the private implementation of this package.


maltezfaria:
Thanks, LGTM.

PR #7, opened 2022-06-22
Stage

vsivadon:
Clean version of the current state of the logger
!! TiledFactorization decoupled !!


maltezfaria:
Thanks. Do you think you can:
- Create a branch with *only* the logger changes
- Make sure the tests are still passing
- Submit a new PR

That would make it much easier to inspect the code.

Issue #5, opened 2022-06-21
Make `TiledFactorization` into its own package + clean-up dependencies


maltezfaria:

maltezfaria:
I think `TiledFactorization` could eventually be a useful package in itself which uses `DataFlowTask`s. Some ideas after the discussion today:

- Add a `TiledMatrix` type which copies the data into contiguous blocks
- Revive the `lu` factorisation
- Do some rigorous performance tests against MKL (and perhaps other specialised libraries?)

maltezfaria:
Closed by #8

PR #8, opened 2022-06-22
extract `TiledFactorization` module

maltezfaria:
- Extract the `TiledFactorization` module into a [package](https://github.com/maltezfaria/TiledFactorization.git)
- Update the dependencies
- Move tests on `cholesky` and `lu` to `TiledFactorization`


maltezfaria:
Partially addresses #5. The other issues mentioned there (careful benchmarking and reviving the `lu` solver) should be moved upstream to the `TiledFactorization` package.

Issue #6, opened 2022-06-21
Improve the API 


maltezfaria:

ffevotte:
FYI: I'm working on something that would look like this:

```
julia> using DataFlowTasks: @dspawn_new

julia> let
           x = rand(1000)
       
           t1 = @dspawn_new begin
               @W x
               sleep(1)
               x .= 0.001
           end
       
           t2 = @dspawn_new sum(@R x)
       
           fetch(t2)
       end
1.0000000000000007
```

where macro calls expand to things like:
```
julia> @macroexpand @dspawn_new begin
           @RW x
           @R  y z
           x .= y .+ z
       end priority=2
quote
    var"#279#t" = (DataFlowTasks.DataFlowTask)((()->begin
                    begin
                        x
                        y
                        x .= y .+ z
                    end
                end), (x, y, z), (DataFlowTasks.READWRITE, DataFlowTasks.READ, DataFlowTasks.READ), 2)                               
    (DataFlowTasks.spawn)(var"#279#t")
end
```

ffevotte:
A side effect of this is that there is no need anymore to import things like `DataFlowTasks.R`, `W` and `RW`. I think is rather nice, but that can cause collisions if there exist `@R`, `@W` or `@RW` macros coming from some user-defined or third-party code.

maltezfaria:
I think the risk of collisions is acceptable -- not too many people define macros :-) 

And I like the proposed `API` !

One thing that would be nice as well is to implement a global way to deactivate the parallelism and execute the code sequentially.

maltezfaria:
closed by #9

PR #9, opened 2022-06-22
New surface API

ffevotte:
This branch addresses issue #6.

All this is still work in progress, but this early PR provides a centralized place to discuss about the new API and its implementation choices.

**TODO:**
- [x] implement the new surface API
- [x] update tests
- [x] make parallelism optional
- [x] update documentation

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -26,45 +26,51 @@ possible. The API revolves around three macros:
 - [`@dasync`](@ref)
 
 They behave like their `Base` counterparts (`@task`, `Threads.@spawn`
-and `@async`), but two additional arguments specifying explicit *data
+and `@async`), but additional annotations specifying explicit *data
 dependencies* are required. The example below shows the most basic usage:
 
 ```@example simple-example
 using DataFlowTasks # hide
-using DataFlowTasks: R,W,RW
 
 A = ones(5)
 B = ones(5)
 d = @dspawn begin
+    @RW A   # A is accessed in READWRITE mode
+    @R  B   # B is accessed in READ mode
     A .= A .+ B
-end (A,B) (RW,R)
+end
 
 fetch(d)
 ```
 
 This creates (and schedules for execution) a `DataFlowTask` `d`
-which access `A` in `READWRITE` mode, and `B` in `READ` mode. The benefit of
+which accesses `A` in `READWRITE` mode, and `B` in `READ` mode. The benefit of
 `DataFlowTask`s comes when you start to compose operations which may mutate the
 same data:
 
 ```@example simple-example
 using DataFlowTasks # hide
-using DataFlowTasks: R,W,RW
 
 n = 100_000
 A = ones(n)
 
 d1 = @dspawn begin
+    @RW A
+
     # in-place work on A
     for i in eachindex(A)
         A[i] = log(A[i]) # A[i] = 0
     end
-end (A,) (RW,)
+end
 
-d2 = @dspawn begin
-    # reduce A
-    sum(A)
-end (A,) (R,)
+# reduce A
+d2 = @dspawn sum(@R A)
+# The above is a shortcut for:
+#   d2 = @dspawn begin
+#       @R A
+#       sum(A)
+#   end
+
 
 c = fetch(d2) # 0
 ```
@@ -84,34 +90,40 @@ outcome is thus always zero.
     an `n` large enough) you will see that you no longer get `0` because `d2`
     may access an element of `A` before it has been replaced by zero!
 
+!!! tip
+    In the `d2` example above, a shortcut syntax was introduced, which
+    allows putting `READ`/`WRITE` annotations directly around arguments in a
+    function call. This is especially useful when the task body is a one-liner.
+
 No parallelism was allowed in the previous example due to a data conflict. To
 see that when parallelism is possible, spawning `DataFlowTask`s will exploit it,
 consider this one last example:
 
 ```@example simple-example
 using DataFlowTasks # hide
-using DataFlowTasks: R,W,RW
 
 n = 100
 A = ones(n)
 
 d1 = @dspawn begin
+    @W A
+
     # write to A
     sleep(1)
     fill!(A,0)
-end (A,) (W,)
+end
 
 d2 = @dspawn begin
+    @R A
+
     # some long computation 
     sleep(5)
     # reduce A
     sum(A)
-end (A,) (R,)
+end
 
-d3 = @dspawn begin
-    # another reduction on A
-    sum(x->sin(x),A)
-end (A,) (R,)
+# another reduction on A
+d3 = @dspawn sum(x->sin(x), @R(A))
 
 t = @elapsed c = fetch(d3)
 
@@ -207,24 +219,30 @@ M2 = CirculantMatrix(copy(v));
 Base.sum(M::CirculantMatrix) = length(M.data)*sum(M.data)
 
 d1 = @dspawn begin
+    @W v
+
     sleep(0.5)
-    println("I write to v")
+    println("d1: I write to v")
     fill!(v,0) 
-end (v,) (W,);
+end;
 
 d2 = @dspawn begin
-    println("I wait for d1 to write")
+    @R M1
+
+    println("d2: I wait for d1 to write")
     sum(M1)
-end (M1,) (R,);
+end;
 
-d3= @dspawn begin
-    println("I don't wait for d1 to write")
+d3 = @dspawn begin
+    @R M2
+
+    println("d3: I don't need to wait for d1 to write")
     sum(M2)
-end (M2,) (R,);
-
-fetch(d2)
+end;
 
 fetch(d3)
+
+fetch(d2)
 ```
 
 ## Scheduler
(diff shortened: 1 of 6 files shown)


ffevotte:
The new API looks like this:

```julia
using DataFlowTasks: @dspawn

let
    x = rand(1000)

    p = 10
    t1 = @dspawn begin
        @W x
        sleep(1)
        x .= 0.001
    end priority=p

    i = 1
    t2 = @dspawn sum(@R x) label="sum($i)"

    result = fetch(t2)

    @show t1.priority   # t1.priority = 10.0
    @show t2.label      # t2.label = "sum(1)"
    @show result        # result = 1.0000000000000007
end
```

maltezfaria:
The syntax looks good to me. I saw that you even included the arrows 🤣 

Do you want me to take a stab at the remaining points? I should be able to handle the "optional" part + the docs and tests. 

> This branch addresses issue #6.
> 
> All this is still work in progress, but this early PR provides a centralized place to discuss about the new API and its implementation choices.
> 
> **TODO:**
> 
> * [x]  implement the new surface API
> * [ ]  make parallelism optional
> * [ ]  update tests
> * [ ]  update documentation

ffevotte:
I already have the tests working... I just need to clean things up a little and will probably commit it tomorrow morning.

As for the "optional" part, I'd rather implement it myself if you don't mind. There are several ways to do it; I'd like to try a few things before I settle on one.

If you want to update the documentation, please don't hesitate to do it. Otherwise, I'll do it once the two other points will have been taken care of.

test/testutils.jl
```diff
@@ -9,11 +9,21 @@ using DataFlowTasks: R,W,RW
 # each block works for `s` seconds
 function fork_join(n,s)
     A = rand(2n)
-    @dspawn do_work(s) (A,) (RW,)
```
ffevotte: Doing things like this is a bit more awkward with the new API

ffevotte:
A priority scheduler test fails when run with multiple threads. I don't think the issue is related to this PR, but maybe you'll have more insight @maltezfaria?

```julia
Fork-join: Test Failed at /home/francois/tmp/DataFlowTasks.jl/test/priorityscheduler_test.jl:22
  Expression: abs(t1 - t2) < 0.01
   Evaluated: 0.6014076429999999 < 0.01
Stacktrace:
 [1] macro expansion
   @ ~/.asdf/installs/julia/1.7.2/share/julia/stdlib/v1.7/Test/src/Test.jl:445 [inlined]
 [2] macro expansion
   @ ~/tmp/DataFlowTasks.jl/test/priorityscheduler_test.jl:22 [inlined]
 [3] macro expansion
   @ ~/.asdf/installs/julia/1.7.2/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
 [4] macro expansion
   @ ~/tmp/DataFlowTasks.jl/test/priorityscheduler_test.jl:15 [inlined]
 [5] macro expansion
   @ ~/.asdf/installs/julia/1.7.2/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
 [6] top-level scope
   @ ~/tmp/DataFlowTasks.jl/test/priorityscheduler_test.jl:14
Test Summary:        | Pass  Fail  Total
Scheduler tests      |    4     1      5
  Julia scheduler    |    4            4
  Priority scheduler |          1      1
    Fork-join        |          1      1
```

test/testutils.jl
maltezfaria: This is a weird and somewhat artificial example since `A` never enters the actual computation, so I think it is fine that the API feels awkward here.

maltezfaria:
I checked out your branch and ran the tests, and they pass on my machine. These tests have a bunch of hidden assumptions (e.g. that the `GC` does not run while the `fork_join` function executes), and are not very robust. I guess we could run `GC.gc()` before each test to make it less likely that the `GC` will run, but the tests still feel brittle to me. I am gonna open a new issue regarding better tests 👍 

In any case, I think this problem has nothing to do with your PR.

p.s. There may be something strange with the `PriorityScheduler` that needs to be investigated, but I think we can push that to the future and simply remove the `otherschedulers.jl` file for now.

maltezfaria:
> I checked out your branch and ran the tests, and they pass on my machine.

Correction: I just ran the test a bunch of times in a row and they pass *most of the time*, but now and then the time difference is significantly larger than the tolerance set there. 

I will take a look at this see if I find the problem; otherwise I suggest we just remove the `PriorityScheduler` until we decide if it is worth having it.

ffevotte:
Regarding the "optional" part (which might be more clearly referred to as a "sequential mode"), I experimented with two ideas:
1. **test during code generation:** `@dspawn compute(...)` checks some global state, and expands to either (when in parallel mode)
    ```julia
    spawn(DataFlowTask(()->compute(...), ...))
   ```
   or simply (in sequential mode)
   ```julia
   compute(...)
   ```

2. **test included in the generated code:** `@dspawn compute(...)` always expands to a code that tests whether sequential mode is active:
   ```julia
   if _sequential_mode()
      compute(...)
   else
      spawn(DataFlowTask(()->compute(...), ...))
   end
   ```

With option 1, we know for sure that supporting "sequential mode" will incur absolutely no runtime overhead, because when sequential mode is deactivated, the generated code is exactly the same as if sequential mode didn't exist at all. However, it's kind of hard to force an update of the code when code generation has already happened once (read: I don't know of any reliable way to make this happen). This means that the switch to "sequential mode" will have to be made before any code is generated. And if `DataFlowTasks` is used in a 3rd-party package that gets precompiled, I'm not sure what a user could expect to happen...

Option 2 feels safer, in that it should always be possible to change `DataFlowTasks`'s sequential mode, even at runtime, without worrying about Julia's internal state being inconsistent w.r.t. the current mode. However, with this option we have to rely **a lot** on compiler optimizations (by "a lot", I mean that if the compiler is not able to optimize the branch away, we even get a type instability).

<br/>

I was originally inclined to favor option 1, but the implications make me uncomfortable, especially when considering what would happen if DataFlowTasks was used as a (potentially indirect) dependency of a 3rd-party package. So I would now tend to favor option 2, especially since my early tests seem to show that the compiler is always able to optimize the branch away.

But what do you think?

maltezfaria:
> But what do you think?

I was thinking about implementing something akin to *option 1*, but I did not think about the implications this would have regarding pre-compilation and issues begin a third-party dependency. In view of that, I think *option 2* is preferable (which is essentially what we already do for other things like logging and the `error_handler`).

Question: can't we have both options, with a flag like `_sequential_mode_generated` which implements option 1, essentially overloading the `_sequential_mode` option, and mention that tweaking with it is highly discouraged due to the reasons you mentioned?

ffevotte:
> Question: can't we have both options?

Why not? We're [greedy](https://julialang.org/blog/2012/02/why-we-created-julia/), right?

I'll see what I can do (probably not before Monday, though).

ffevotte:
With the latest commit:

```julia
julia> using DataFlowTasks

julia> x = rand(10);
julia> foo(x) = @dspawn sum(@R x)
foo (generic function with 1 method)

julia> t = foo(x)      # Parallelism is active by default => this returns a DataFlowTask
DataFlowTask (done) 1

julia> fetch(t)
4.811620110814342

julia> DataFlowTasks.enable_sequential(true);   # Parallelism is now disabled
julia> foo(x)                                   # => this directly returns a Float
4.811620110814342

# Although the choice between sequential & parallel modes is done (by default)
# after code generation, the return type is inferrable
julia> @code_warntype foo(x)
MethodInstance for foo(::Vector{Float64})
  from foo(x) in Main at REPL[3]:1
Arguments
  #self#::Core.Const(foo)
  x::Vector{Float64}
Locals
  #1::Union{}
  @_4::Int64
  ##task#291::Union{}
  par::Symbol
  dyn::Symbol
Body::Float64
[...]
```

There is no automated test yet

ffevotte:
Now that test and docs are up-to-date, I think this can be reviewed. And maybe re-based on the latest master?

docs/src/index.md
```diff
@@ -207,24 +219,30 @@ M2 = CirculantMatrix(copy(v));
...
 
-d3= @dspawn begin
-    println("I don't wait for d1 to write")
-    sum(M2)
-end (M2,) (R,);
+d3 = @dspawn begin
+    @R M2
 
-fetch(d2)
+    println("d3: I don't need to wait for d1 to write")
```
ffevotte: I don't know why, but this output line never shows up in the documentation.

maltezfaria:
I will try to go over the changes later today and let you know if I have any questions, thanks!

docs/src/index.md
maltezfaria: I don't know either... perhaps printing to the repl asynchronously from various tasks is not a very robust thing? In any case, I think using "print" as a way to indicate order was not very good, so I simplified the example a bit.

src/dataflowtask.jl
```diff
@@ -137,22 +137,140 @@ function memory_overlap(di,dj)
...
 """
-macro dtask(expr,data,mode,p=0,label="")
-    :(DataFlowTask(
-        ()->$(esc(expr)),
-        $(esc(data)),
-        $(esc(mode)),
-        $(esc(p)),
-        $(esc(label))
-        )
+function enable_sequential(seq::Bool = true; static::Bool = false)
```
maltezfaria: What about renaming this to `force_sequential`?

maltezfaria:
I took the liberty to implement the minor comments I had and push to your branch. I will rebase this soon and proceed with the merge.

PR #10, opened 2022-06-23
New version of logger

maltezfaria:
Contains most of the changes by @Vincent-Sivadon, including:
- More complex/rich recipes for plotting
- Thread-safe logging
- Mechanism to globally enable/disable logging

Still missing:
- [x] Log the time to insert nodes in the `DAG`
- [x] Use the *a posteriori* critical path as a reference optimal time when plotting
- [ ] Testing
- [ ] Documentation

### DIFF
--- a/src/dag.jl
+++ b/src/dag.jl
@@ -226,3 +226,77 @@ function Base.show(io::IO, dag::DAG{T}) where {T}
     s2 = e==1 ? "" : "s"
     print(io, typeof(dag)," with $n node$s1 and $e edge$s2 (capacity of $(dag.sz_max[]) nodes)")
 end
+
+
+############################################################################
+#                           Critical Path                                  
+############################################################################
+
+"""
+    notvisited_min(dist, visited) --> minnode 
+Extract the index minnode of the minimum value of dist
+such that visited[minnode] = false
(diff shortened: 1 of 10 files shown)


src/dag.jl
```diff
@@ -227,21 +234,76 @@ function Base.show(io::IO, dag::DAG{T}) where {T}
...
+    push!(path, currnode)
+
+    # Run through previous[] to get the path
+    while currnode != source
+        currnode = previous[currnode]
+        push!(path, currnode)
+    end
+
+    path
 end
```
maltezfaria, line 309: We should avoid the creation of a dense adjacency matrix for doing this computation. Also, I think we could do something simpler than Djisktra here (see e.g. https://www.geeksforgeeks.org/find-longest-path-directed-acyclic-graph/)

maltezfaria:
I am gonna merge this so that we can test some things on main, but I have opened the following tickets which need to be addressed in a subsequent PR:

1. The plot recipes are brittle and inconsistent. I am not sure how to fix this though. Perhaps do not use `Plots`? (#15)
2. The longest path algorithm is not very efficient, and its implementation could be made more generic (#14)

PR #17, opened 2022-07-01
Test TiledFactorization in CI

ffevotte:
With this PR, the CI workflow integrates new steps during which [https://github.com/maltezfaria/TiledFactorization](TileFactorization) is cloned and its automated tests are passed. One advantage of integrating this as new steps in the existing workflow is that code coverage statistics should account for the code paths activated by TiledFactorization.

FWIW, an alternate way of doing things would have been to define a new, dedicated job (alongside the "Julia" test matrix and the "Documentation" job). This could make it easier to see whether failures come from internal tests or TiledFactorization, but it would not easily allow code coverage statistics to include TiledFactorization tests.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,6 +33,14 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+      - run: git clone https://github.com/maltezfaria/TiledFactorization
+      - run: |
+          julia --project="TiledFactorization" -e '
+            import Pkg
+            Pkg.develop(Pkg.PackageSpec(path=pwd()))
+            Pkg.instantiate()
+            Pkg.test(coverage=true)
+          '
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v1
         with:


ffevotte:
To expand on the code coverage topic, it looks like `TiledFactorization` tests `src/arrayinterface.jl` a bit more extensively than the internal tests in `DataFlowTasks`:

https://app.codecov.io/gh/maltezfaria/DataFlowTasks.jl/commit/656d2a164cec9c990316c52f54232b2ae61df473

ffevotte:
I also wanted to say that `TiledFactorization` tests tend to take a *very* long time to pass (although they are accasionally shorter). I suspect compilation-induced latency to be the culprit here, but I haven't investigated much.

maltezfaria:
> To expand on the code coverage topic, it looks like `TiledFactorization` tests `src/arrayinterface.jl` a bit more extensively than the internal tests in `DataFlowTasks`:
> 
> https://app.codecov.io/gh/maltezfaria/DataFlowTasks.jl/commit/656d2a164cec9c990316c52f54232b2ae61df473

That makes sense as `TiledFactorization` makes extensive use of the `Array` interface. It is very nice to have the test and test coverage automated -- thanks!

maltezfaria:
> I also wanted to say that `TiledFactorization` tests tend to take a _very_ long time to pass (although they are accasionally shorter). I suspect compilation-induced latency to be the culprit here, but I haven't investigated much.

I think you are correct -- the culprit is the compilation-induced latency. To be more precise, I think the issue comes from the use of `@turbo` from `TiledFactorization` to generate some efficient kernels for the small sized `_chol` and `_lu` methods. 

**Note**: After seeing that most of the time is actually spent doing the Schur complement, which is delegated to `Octavian`, I wonder if we could not simply use the generic `ldiv` and `chol` methods for the `LinearAlgebra`. That would significantly simply the package, reduce the number of lines of code, and drop a few "heavy" dependencies. Perhaps worth testing?

PR #18, opened 2022-07-01
Makie Scene

### LINKED ISSUE #15: Plot recipes
The plot recipes are not robust, and depend largely on the backend used. Also, sometimes the label are incorrectly placed, leading to confusion when reading the portion of time spent.

vsivadon:
Plotting through Makie environnement.

- [x] Interactive trace plot : over on a task gives informations on it
- [x] Adaptative labeled DAG plotting with GraphViz
- [x] Task repartition
- [x] Infinite Proc time comparaison
- [x] Fix the current tests
- [x] Use `Requires` to make `Makie` optional
- [x] Add examples to documentation
- [x] Add some tests to the logger utilities
- [x] Measure `GC` time during task insertion
- [x] Add support for regex expression when parsing the `categories` label


src/dag.jl
```diff
@@ -31,6 +31,7 @@ function DAG{T}(sz = typemax(Int)) where T
     cond_push  = Condition()
     cond_empty = Condition()
     _buffer    = Set{Int}()
+    sz == Inf || sizehint!(_buffer, sz)
```
ffevotte: @maltezfaria Is this still needed?

Note that because of line 26, `sz` is never `Inf` but could be `typemax(Int)`. And I would expect `sizehint!(_buffer, typemax(Int))` to exhaust memory...

maltezfaria: @ffevotte you are right, this is neither needed nor correct. @Vincent-Sivadon can you fix this in the PR?

maltezfaria:
@Vincent-Sivadon is this ready for review?

vsivadon:
@maltezfaria I guess i'm not sure about the "interactive plot documentation" with jldocstring (couldn't find a way to check if it worked), and also : should GLMakie and GraphViz be conditional packages with Requires or just GLMakie ? I think i'll change to juste GLMakie.
Apart from that the core code is ready for review.

ffevotte:
The two commits above:
1. make sure the logging tests are actually run during CI tests (the file was originally not listed in `runtests.jl`)
2. avoid warnings about optional dependencies (such as `GLMakie`) being used in the source file but not listed in `Project.toml`. The way to do this is documented in `Requires.jl`'s [README](https://github.com/JuliaPackaging/Requires.jl):
   > In the `@require` block, or any included files, you can use or import the package, but note that you must use the syntax `using .Gadfly` or `import .Gadfly`, rather than the usual syntax. Otherwise you will get a warning about `Gadfly` not being in dependencies.

ffevotte:
> i'm not sure about the "interactive plot documentation" with jldocstring (couldn't find a way to check if it worked),

I think the doctests for `plot` did not run because the optional dependencies were not loaded in the documentation environment. If I'm not mistaken the commit above should take care of that (i.e. doctests will run during CI, but I expect them to fail).

**EDIT:** as expected, doctests now run during CI, but fail. A first cause of failure was related to CI running on headless servers. It should now be fixed thanks to commit 5a80794. The remaining failures are related to `jldoctest` not following the required format. This is reproducible when trying to generate the documentation locally.

src/DataFlowTasks.jl
```diff
@@ -11,6 +11,7 @@ using Logging
 using ThreadPools
 using DataStructures
 using RecipesBase
```
ffevotte: Is the dependency on `RecipesBase` still required?

src/plot.jl
```diff
@@ -0,0 +1,448 @@
+@info "Loading DataFlowTasks plot utilities"
+
+using .GLMakie
+using Cairo
+using FileIO  # to be made conditionnal
```
ffevotte: I think both of these should be made conditional

src/plot.jl
```diff
@@ -0,0 +1,448 @@
+@info "Loading DataFlowTasks plot utilities"
+
+using .GLMakie
```
ffevotte: Why is there a `using .GLMakie` but no `using .GraphViz`?
The `@require` statement seems to indicate that both are needed for this to work...

test/logger_test.jl
```diff
@@ -0,0 +1,40 @@
...
+
+A = ones(20, 20)
+B = ones(20, 20)
+
+work(copy(A), copy(B))
+
+DataFlowTasks.resetlogger!()
+DataFlowTasks.TASKCOUNTER[] = 0
+
+GC.gc()
```
ffevotte: Is this needed for the tests to pass, or only to get prettier results?

test/logger_test.jl
```diff
@@ -0,0 +1,40 @@
...
+
+logger = DataFlowTasks.getlogger()
+
+@test length(logger.tasklogs) == Threads.nthreads()
+nbtasks = DataFlowTasks.nbtasknodes(logger)
+nbinsertion = sum(length(insertionlog) for insertionlog ∈ logger.insertionlogs)
+@test nbtasks == 6
+@test nbinsertion == 6
+
+path = DataFlowTasks.criticalpath(logger)
```
ffevotte, line 38: I get `path = [7, 5, 3, 1]` on my machine. Is that expected? I would have thought the critical path to be of size 3.

In any case, it would be better to check that the results are correct. As it stands now, we only check that the code runs without error.

test/logger_test.jl
```diff
@@ -0,0 +1,40 @@
...
+dotstr = DataFlowTasks.loggertodot(logger)
```
ffevotte, line 42: When I extract the results and generate the corresponding DAG using graphviz/dot, it looks like this is correct. However we don't currently check anything about this during CI.

This is more difficult to check than `path` above, but could we at least look for a few sensible substrings in the generated dot format?

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+
+"""
+    plot(logger; categories)  
+Plot DataFlowTasks `logger` labeled informations with categories.  
+Note : if there's too many tasks, the DAG won't be plotted,
+although you can manually call `dagplot(logger)`.
+
+## Example  
+Use the above exemple with
+```jldoctest
```
ffevotte: In order for this to be a fully functional doctest, it needs to meet the required format. Either it should be formatted as a REPL session:
```
julia> 1+1
2
```

OR it should include an `# output` header at the end, which `Documenter` will compare to the actual output in order to determine whether the test fails or not:
```
1+1
# output
2
```

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+using GLMakie, GraphViz
+using DataFlowTasks
+using DataFlowTasks: RW
+import DataFlowTasks as DFT
```
ffevotte: This might be a matter of personnal preference, but I don't really like such import statements and would like to avoid advertising their use here. Couldn't we simply do something like the following?
```
using DataFlowTasks: plot, getlogger
...
plot(getlogger(), categories=["A", "B"])
```

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+
+computing(A) = exp.(sum(A).^2).^2
+function work(A, B)
+    @dspawn computing(@RW(A)) label="A"
+    @dspawn computing(@RW(B)) label="B"
+    @dspawn computing(@RW(A)) label="A"
+    @dspawn computing(@RW(B)) label="B"
```
ffevotte: Shouldn't we put more complex labels here, such as `A(1)`, `A(2)`, etc? This would advertize the feature about filtering on labels.

Also, I'm wondering if this example isn't a bit too contrived: none of the specified data dependencies corresponds to what the code actually does, and the labels are not too descriptive either. Maybe we could keep something along the same lines, but having an initialization task (that is `@W`) and a mutation task (`@RW`) for each data. They could be labelled following a pattern such as `A(init)` or `A(mutate)`.

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+end
+
+# Context
+A = ones(1000, 1000)
+B = ones(1000, 1000)
+
+# Compilation
+work(copy(A), copy(B))
```
ffevotte: If this is a recommendation for users to avoid seeing compilation- or GC-related artifacts, this is really nice! But maybe it could be made a little more explicit by adding a few comments.

You can get inspiration from the documentation of `Profile` or `ProfileView`, but maybe something along the lines of
```
# Compilation
# run your code once to avoid seeing artifacts related to compilation in your logged data

File: src/plot.jl [hunk 12]
```diff
@@ -0,0 +1,448 @@
...
+
+# Reset logger and taskcounter
+DFT.resetlogger!()
+DFT.TASKCOUNTER[] = 0
```
ffevotte: Why two lines? Couldn't both these operations be realized by `resetlogger!`?

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+    @dspawn computing(@RW(B)) label="B"
+end
+
+# Context
+A = ones(1000, 1000)
+B = ones(1000, 1000)
+
+# Compilation
+work(copy(A), copy(B))
+
```
ffevotte: Maybe add a call to the GC here (with an explanation as well) if this is something users should be encouraged to do when profiling real code.

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+DFT.TASKCOUNTER[] = 0
+
+# Reak Work
+work(A, B)
+
+# Logger Visualization
+DFT.plot(DFT.getlogger(), categories=["A", "B"])
+```
+"""
+function plot(logger::Logger; categories=String[])
```
ffevotte: Could we put `getlogger()` as a default value for `logger`? I would expect this is how it will be used 99% of the time.

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+    # Events management
+    react(axtrc, logger, gantt)
+
+    # Terminal Informations
+    # ---------------------
+    @info "Computing    : $(loginfo.computingtime)"
+    @info "Inserting    : $(loginfo.insertingtime)"
+    @info "Other        : $(loginfo.othertime)"
+
+    fig
```
ffevotte: It's a good idea to return `fig`. It means that in automated tests, we could do something like:
```
fig = plot(getlogger(), categories=["A", "B"])
save("logger.png", fig)
```

At least when this runs on a local machine, we can check the results. And the generated figure could be included in the documentation.

ffevotte (commented):
Here is a first list of questions/comments that I feel should be addressed before merging.

src/plot.jl
```diff
@@ -0,0 +1,448 @@
...
+Plot DataFlowTasks `logger` labeled informations with categories.  
+Note : if there's too many tasks, the DAG won't be plotted,
+although you can manually call `dagplot(logger)`.
+
+## Example  
+Use the above exemple with
+```jldoctest
+using GLMakie, GraphViz
+using DataFlowTasks
+using DataFlowTasks: RW
```
ffevotte: I don't think `using DataFlowTasks: RW` is needed anymore

src/plot.jl
vsivadon: Because we had to call "using GraphViz" before compiling this code, using GraphViz.funcname was sufficient, so it didn't complained.
I added the using .GraphViz for consistency.

test/logger_test.jl
vsivadon: Indeed, i'm removing the work on copy. I'll let the reset calls, in case DataFlowTasks was called before this test (don't really know if it can happen in the CI context though).

test/logger_test.jl
vsivadon: Yes it's expected : for the critical path, i add a ghost node (1) that represents the beginning. 1 is connected with all tasks without in-neighbours by a edge of length 0.
I'm taking this into account when using the result. I should in fact do a treatment before sending the result.

The path for this "work" is not determinist, we can't predict what path (1->3->5 or 2->4->6) will be the fastest.

I will write another predictible test.

src/plot.jl
vsivadon: I did realise that afterwards (wednesday). It's the first time i use this documentation system, it was a first draft, i'll work on that.

src/plot.jl
vsivadon: Good idea ! I'll write that.

src/plot.jl
vsivadon: I agree, i did add it on my local version (when experimenting on the doc).

src/plot.jl
Vincent-Sivadon: > The two commits above:
> 
> 1. make sure the logging tests are actually run during CI tests (the file was originally not listed in `runtests.jl`)
> 2. avoid warnings about optional dependencies (such as `GLMakie`) being used in the source file but not listed in `Project.toml`. The way to do this is documented in `Requires.jl`'s [README](https://github.com/JuliaPackaging/Requires.jl):
>    > In the `@require` block, or any included files, you can use or import the package, but note that you must use the syntax `using .Gadfly` or `import .Gadfly`, rather than the usual syntax. Otherwise you will get a warning about `Gadfly` not being in dependencies.

Thanks for the tips, i planned on asking about that (didn't pay attention to this part of the Requires.jl doc it seems)

vsivadon:
> > i'm not sure about the "interactive plot documentation" with jldocstring (couldn't find a way to check if it worked),
> 
> I think the doctests for `plot` did not run because the optional dependencies were not loaded in the documentation environment. If I'm not mistaken the commit above should take care of that (i.e. doctests will run during CI, but I expect them to fail).
> 
> **EDIT:** as expected, doctests now run during CI, but fail. A first cause of failure was related to CI running on headless servers. It should now be fixed thanks to commit [5a80794](https://github.com/maltezfaria/DataFlowTasks.jl/commit/5a807942111edd290458a6022aab580a558679e0). The remaining failures are related to `jldoctest` not following the required format. This is reproducible when trying to generate the documentation locally.

I have experimented quite a lot on this documentation problem wednesday, but on the "make.jl" file side (in vain). Thanks again for the fix, i'll take care of the remaining problems.

vsivadon:
> The two commits above:
> 
> 1. make sure the logging tests are actually run during CI tests (the file was originally not listed in `runtests.jl`)
> 2. avoid warnings about optional dependencies (such as `GLMakie`) being used in the source file but not listed in `Project.toml`. The way to do this is documented in `Requires.jl`'s [README](https://github.com/JuliaPackaging/Requires.jl):
>    > In the `@require` block, or any included files, you can use or import the package, but note that you must use the syntax `using .Gadfly` or `import .Gadfly`, rather than the usual syntax. Otherwise you will get a warning about `Gadfly` not being in dependencies.

Thanks for the tips, i planned on asking about that (didn't pay attention to this part of the Requires.jl doc it seems)

vsivadon:
@ffevotte Despite the changes you made to the CI doc script, i still have this issue (even while the jldocstring code block is removed) while building the documentation. Did you also have this error message ? Or is it just on my side ?
image

ffevotte:
No, I don't see anything like that. Here is what I get with the `jldoctest` removed (and once the `docs` subproject has been correctly setup):
```
shell$ cd DataFlowTasks.jl/docs
shell$ julia --project -e '\
          using Documenter: DocMeta, doctest; \
          using GLMakie, GraphViz; \
          using DataFlowTasks; \
          DocMeta.setdocmeta!(DataFlowTasks, :DocTestSetup, :(using DataFlowTasks); recursive=true); \
          doctest(DataFlowTasks)'
[ Info: Loading DataFlowTasks plot utilities
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: Skipped ExpandTemplates step (doctest only).
[ Info: Skipped CrossReferences step (doctest only).
[ Info: Skipped CheckDocument step (doctest only).
[ Info: Skipped Populate step (doctest only).
[ Info: Skipped RenderDocument step (doctest only).
Test Summary:           | Pass  Total
Doctests: DataFlowTasks |    1      1
```

(FYI: the command above can be copy-pasted from `.github/workflows/CI.yml`. The `docs` subproject can also be set up by copy-pasting the relevant command from the CI file. But that does not seem to be your problem here, since `Documenter` at least started doing its job)

src/plot.jl
ffevotte: OK. Now that this docstring is a simple example (not a doctest anymore), would it make sense to add a call to `plot` in one of the "real" test cases. I'm not sure we can test that the result is correct, but it might be nice to at least making sure the code is runnable without errors?

src/plot.jl
```diff
@@ -0,0 +1,432 @@
...
+
+## Example  
+Use the above exemple with
+```@example
+using GLMakie, GraphViz, Cairo, FileIO
+using DataFlowTasks
+using DataFlowTasks: plot, resetlogger!, sync
+
+init!(A) = (A .= rand())               # Write
+mutate!(A) = (A = exp.(sum(A).^2).^2)  # Read/Write
```
ffevotte: Shouldn't this be `A .= exp.(...)` instead?

src/plot.jl
ffevotte: In relation to the doctest conversation above, I'm wondering whether it would make sense to amend the test in `logger_test.jl` to call `plot` and `savefig`.
During CI, this would at least ensure that the code runs. When running the tests locally, we could then look at the generated figure to check that it looks good. Would that make sense?

ffevotte:
The commit above adds support for `String => Regex` pairs when defining categories for plotting profiling information.

With the example from the docstring:
```julia
function work(A, B)
    @dspawn init!(@W(A))      label="init A"
    @dspawn init!(@W(B))      label="init B"
    @dspawn mutate!(@RW(A))   label="mutate A"
    @dspawn mutate!(@RW(B))   label="mutate B"
    @dspawn get(@R(A), @R(B)) label="read A,B"
    sync()
end

[...]

plot(categories=[
    "init",               # matches "init A" and "init B", equivalent to: "init" => r"init"
    "read",               # matches "read A,B"
    "work on B" => r"B\$" # matches "mutate B" (could match "init B" and "read A,B" too, but these tasks were already matched by previous labels
    # task "mutate A" is left uncategorized
])
```

.github/workflows/CI.yml
```diff
@@ -45,14 +45,20 @@ jobs:
...
           version: '1'
+      - run: sudo apt-get update && sudo apt-get install -y xorg-dev mesa-utils xvfb libgl1 freeglut3-dev libxrandr-dev libxinerama-dev libxcursor-dev libxi-dev libxext-dev
       - run: |
           julia --project=docs -e '
             using Pkg
             Pkg.develop(PackageSpec(path=pwd()))
+            Pkg.add(PackageSpec("GLMakie"))
+            Pkg.add(PackageSpec("GraphViz"))
+            Pkg.add(PackageSpec("Cairo"))
+            Pkg.add(PackageSpec("FileIO"))
```
maltezfaria: Why can't these dependencies be declared on the `Project.toml` of the `docs`? I think I missed something in the previous discussions.

ffevotte: You're absolutely right, there's no reason not to add these to the `docs` environment.

More generally, I think we should make it easy for developers using `DataFlowTasks` to install these optional deps in a clean way (maybe not in their global environment). If we do this, the `docs` environment (and maybe `test` too) could benefit from that.

Laurent & I were thinking about making a subdirectory somewhere defining a proper environment for these optional deps. We should discuss this today

maltezfaria:
`main` has now been merged into this branch, making it ready for the merge once the reviews are finished. I am gonna go ahead and bump it to *ready for review*.

vsivadon:
I wrote the small changes suggested above. Although i didn't write the visualization test, because i'm not sure what to do : do we need to call the plot function in the test files so the CI runs it, or is the documentation jldoctest enough ?

PR #20, opened 2022-07-07
Allow access mode tags in macro parameters + more macro tests

ffevotte:
I had an idea regarding the "awkwardness" of call like

```julia
@task begin
   @RW x
   @R  y
   fun(x, adjoint(y))
end
```

With this PR, it is possible to specify access modes after the task body, much like the old API worked:

```julia
@task fun(@RW(x), adjoint(y)) @R(y)
```

This means that the list of optional arguments to the macro can now be a mix of parameters specifications (like `priority` & `label`) and access mode specs. In other words, the `@dspawn` macro can now be invoked in even more ways than before, so I felt like more tests were in order. This PR also tries to check the correctness of macro expansion more extensively, by testing all ways to specify data access modes and task parameters.

If you like it, I can document the new syntax. And if we merge it in the main branch, we can use this syntax in `TiledFactorization`.

### DIFF
--- a/src/dataflowtask.jl
+++ b/src/dataflowtask.jl
@@ -180,37 +180,45 @@ function _dtask(continuation, expr::Expr, kwargs; source=LineNumberNode(@__LINE_
     data = []
     mode = []
 
-    # Register access mode `m` for all data listed in `expr`
-    # If multiple data are listed, only return the first one
-    function register_modes(expr, m)
-        for i in 3:length(expr.args)
-            push!(data, expr.args[i])
-            push!(mode, m)
-        end
-        return expr.args[3]
-    end
-
-    # Detect @R/@W/@RW tags in the task body:
-    # - register the associated data and mode
-    # - remove tags from the final expression
-    transform(x) = x
-    function transform(x::Expr)
-        if x.head == :macrocall
+    # Try to parse `expr` as an access tag applied to some data
+    # - if not successful: return `nothing`
+    # - otherwise: register data access
+    try_register_access(expr) = nothing
+    function try_register_access(expr::Expr)
+        if expr.head == :macrocall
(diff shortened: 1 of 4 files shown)


maltezfaria:
> I had an idea regarding the "awkwardness" of call like
> 
> ```julia
> @task begin
>    @RW x
>    @R  y
>    fun(x, adjoint(y))
> end
> ```
> 
> With this PR, it is possible to specify access modes after the task body, much like the old API worked:
> 
> ```julia
> @task fun(@RW(x), adjoint(y)) @R(y)
> ```
> 
> This means that the list of optional arguments to the macro can now be a mix of parameters specifications (like `priority` & `label`) and access mode specs. In other words, the `@dspawn` macro can now be invoked in even more ways than before, so I felt like more tests were in order. This PR also tries to check the correctness of macro expansion more extensively, by testing all ways to specify data access modes and task parameters.
> 
> If you like it, I can document the new syntax. And if we merge it in the main branch, we can use this syntax in `TiledFactorization`.

I like the idea as it allows fo the "one-line" syntax that we had before, while keeping the cool new stuff 😃 

I went over the files and the tests, and they look good to me. Can I go ahead and merge?

ffevotte:
> I went over the files and the tests, and they look good to me. Can I go ahead and merge?

Yes, unless you'd like me to add some documentation (but we can do it later as well)

ffevotte:
I went ahead and added at least some details in the `@dspawn` docstring.

maltezfaria:
> I went ahead and added at least some details in the `@dspawn` docstring.

Thanks. I am gonna go ahead and merge this then. The goal as far the docs go is to do some sort of *case study* where we showcase e.g. the `TiledFactorization` usage. We can then cover the various facets of the API.

PR #22, opened 2022-08-16
Documentation

ffevotte:
The documentation effort made by @Vincent-Sivadon is still under way, but I'm creating this WIP PR so that we have a space to discuss issues related to it.


ffevotte:
One issue that was brought up in earlier discussions comes from the use of a halo or zeros in the Gauss-Seidel example. IIUC, this is currently implemented by using a modified version of `PseudoTiledMatrix` that is aware of the existence of a halo, but in which only the "interior" of the matrix is tiled. This makes it possible to make the algorithm loops completely unaware of the existence of a halo. In particular, all tiles can assume that they have neighboring data in all directions.

OTOH, an other approach would consist in making the matrix itself unaware of the halo. The algorithm itself should therefore take care of iterating only on valid indices. In terms of documentation, this would have the added benefit that boundary tiles have less than 4 neighbors, so that the number of data dependencies is not always the same. It would be interesting to see whether this can be handled in a friendly way by DataFlowTasks and, if so, to use Gauss Seidel as an example to show how to.

A first idea that comes to mind to declare dependencies that don't always exist would be something along those lines:
```
@R (i>2 ? A[i-1,:] : nothing)
```
But I'm not sure whether this already works out of the box from a syntax viewpoint. Maybe it also requires implementing new methods like
```
memory_overlap(::Nothing, ::Any) = false
```

README.md
```diff
@@ -7,3 +7,194 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+  @RW A
+  readwrite!(A)
+end
+
+DataFlowTasks.sync()
+```
+
+This will generate the DAG (Directed Acyclic Graph) below that represents the dependencies between tasks. This means that the task 2 and 3 can be run in parallel. We see how it's the memory that matters here.
+
+![Basic](graph.png)
```
maltezfaria: Are you sure that this is the `DAG` generated? I seems like tasks 2 and 3 operate on new data, and not on a view of `A`, so I would suspect that they are independent of everything else.

maltezfaria:
> One issue that was brought up in earlier discussions comes from the use of a halo or zeros in the Gauss-Seidel example. IIUC, this is currently implemented by using a modified version of `PseudoTiledMatrix` that is aware of the existence of a halo, but in which only the "interior" of the matrix is tiled. This makes it possible to make the algorithm loops completely unaware of the existence of a halo. In particular, all tiles can assume that they have neighboring data in all directions.
> 
> OTOH, an other approach would consist in making the matrix itself unaware of the halo. The algorithm itself should therefore take care of iterating only on valid indices. In terms of documentation, this would have the added benefit that boundary tiles have less than 4 neighbors, so that the number of data dependencies is not always the same. It would be interesting to see whether this can be handled in a friendly way by DataFlowTasks and, if so, to use Gauss Seidel as an example to show how to.
> 
> A first idea that comes to mind to declare dependencies that don't always exist would be something along those lines:
> 
> ```
> @R (i>2 ? A[i-1,:] : nothing)
> ```
> 
> But I'm not sure whether this already works out of the box from a syntax viewpoint. Maybe it also requires implementing new methods like
> 
> ```
> memory_overlap(::Nothing, ::Any) = false
> ```

I am lost here, can you guys give me a reference to what you are calling the Gauss-Seidel algorithm? I can't reconcile what I have in mind, which is the iterative method, to this *five-point stencil* description that I found on the docs 😕

ffevotte:
> I am lost here, can you guys give me a reference to what you are calling the Gauss-Seidel algorithm? I can't reconcile what I have in mind, which is the iterative method, to this five-point stencil description that I found on the docs confused

Yes, I couldn't agree more. I should have mentioned that this is also a question I asked Vincent yesterday. @Vincent-Sivadon seems to have found this terminology in a paper on dataflow parallelism that I haven't seen yet...

README.md
vsivadon: It isn't indeed, using views does work. It's related to the issue mentioned yesterday. I forgot about the copying happenning. As François said a while ago, we should make it clear in the doc to avoid those problems.

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+
+DataFlowTasks.reset!()
+write!(X, alpha) = (X .= alpha)
+readwrite!(X) = (X .+= 2)
+n = 1000
+A = ones(n)
+@dspawn readwrite!(@RW(A))              # 1
+@dspawn write!(@R(@view A[1:500]), 2)   # 2
+@dspawn write!(@R(@view A[501:n]), 3)   # 3
+@dspawn readwrite!(@RW(A))              # 4
```
ffevotte: Shouldn't this be:
```
@dspawn readwrite!(@RW(A))              # 1
@dspawn write!(@W(@view A[1:500]), 2)   # 2
@dspawn write!(@W(@view A[501:n]), 3)   # 3
@dspawn readwrite!(@RW(A))              # 4
```

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+![Cholesky_Image](docs/src/Cholesky_2ndStep.png)
+
+The code of the sequential yet tiled factorization algorithm will be :
+
+```julia
+tilerange(ti, ts) = (ti-1)*ts+1:ti*ts
+function cholesky_dft!(A, ts)
+    m,n = size(A)
+    n%ts != 0 && error("Tilesize doesn't fit the matrix")
+    tn = round(Int, n/ts)
```
ffevotte: Under the constraint that `n%ts==0`, the computation of `tn` could be simplified to `tn = n÷ts`

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+So we have 3 types of tasks : the Cholesky factorization (I), the triangular solve (II), and the schur complement (III).  
+If we have a matrix A decomposed in `n x n` tiles, then the algorithm will have `n` steps. It implies that the step `i ∈ [1:n]` do `1` time (I), `(i-1)` times (II), and `(i-1)²` times (III). We illustrate the 2nd step of the algorithm in the following image.
+
+![Cholesky_Image](docs/src/Cholesky_2ndStep.png)
+
+The code of the sequential yet tiled factorization algorithm will be :
+
+```julia
+tilerange(ti, ts) = (ti-1)*ts+1:ti*ts
+function cholesky_dft!(A, ts)
```
ffevotte: maybe name this version `cholesky_tiled!`

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+        end
+
+        # Submatrix update (III)
+        for tj in ti+1:tn
+            for tk in tj:tn
+                rj = tilerange(tj, ts)  ;  rk = tilerange(tk, ts)
+                mul!(view(A,rj,rk), adjoint(view(A,ri,rj)), view(A,ri,rk))
+            end
+        end
+    end
```
ffevotte: Couldn't all the views manipulations here be simplified if we defined once and for all something like:
```
Atile = [view(A, tilerange(ti, ts), tilerange(tj, ts)) for ti in 1:tn, tj in 1:tn]
```
Then, `view(A, rj, rk)` can be simplified to `Atile[tj,tk]`

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+        for tj in ti+1:tn
+            for tk in tj:tn
+                rj = tilerange(tj)  ;  rk = tilerange(tk)
+                @dspawn mul!(@RW(view(A,rj,rk)), @R(adjoint(view(A,ri,rj))), @R(view(A,ri,rk))) label="mul ($tj,$tk)"
+            end
+        end
+    end
+    DataFlowTasks.sync()
+    # Construct the factorized object
+    return Cholesky(A,'U',zero(LinearAlgebra.BlasInt))
```
ffevotte: Broader question here: aren't these lines morally equivalent to the following?
```
result = @dspawn Cholesky(@R(A), 'U', zero(LinearAlgebra.BlasInt))
return fetch(result)  # this is where synchronization happens implicitly
```
If so, I would tend to favour this latter formulation, because I think it is more composable: we only ensure that all computations modifying `A` are done. If this was inserted in a larger algorithm, we could possibly have other parts of the computation involving `DataFlowTask` without interfering with each other.

In other words, I kinda feel like `DataFlowTasks.sync()` potentially has wider side-effects than we probably want, so its use should perhaps not be encouraged too much (unless it is unavoidable, of course)

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+A = ones(n)
+@dspawn readwrite!(@RW(A))              # 1
+@dspawn write!(@R(@view A[1:500]), 2)   # 2
+@dspawn write!(@R(@view A[501:n]), 3)   # 3
+@dspawn readwrite!(@RW(A))              # 4
+DataFlowTasks.sync()
+DataFlowTasks.dagplot()
+```
+
+This will generate the DAG (Directed Acyclic Graph) above that represents the dependencies between tasks. This means that the task 2 and 3 can be run in parallel. We see how it's the memory that matters here.
```
ffevotte: I don't see any DAG image in the doc?

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+
+# Context
+n  = 4096
+ts = 512
+A = rand(n, n)
+A = (A + adjoint(A))/2
+A = A + n*I
+
+# Compilation
+cholesky_dft!(copy(A), ts)
```
ffevotte: We should probably add one or two lines of code here to check whether the computation is correct

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+
+# Reset environnement
+DFT.resetlogger!()
+GC.gc()
+
+# Real work to be analysed
+cholesky_dft!(A ,ts)
+
+# Plot
+DFT.plot(categories=["chol", "ldiv", "mul"])
```
ffevotte: The labels don't seem to be in accordance to the image shown below.

Also, now that the dag is plotted separately, maybe we should add the line to generate it.

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+## Profiling
+
+We can illustrate the parallelization implied by those modifications. `DataFlowTasks.jl` comes with 2 main profiling tools whose outputs for the case presented above, with a matrix of size (2000, 2000) divided in blocks of (500, 500), are as follows :
+
+![Trace Plot](example.png)
+![Dag Plot](exampledag.svg)
+
+We'll cover in details the usage and possibilities of the visualization in the documentation.
+
+Note that the visualization tools are not loaded by default, it requires a `Makie` backend and/or `GraphViz` loaded in the REPL. It's meant to be used in development, so it won't pollute the environment you want to use DFT in.
```
ffevotte: Yep! If/when #26 (or something like it) gets merged, we'll have to update the doc here accordingly.

README.md
ffevotte: I would also enclose all these calls in a block, otherwise if people copy-paste them one at a time each task will be long finished when the following task is spawned.

README.md
ffevotte: And also perhaps the lines illustrating profiling should be put in a separate block in the section below.

README.md
```diff
@@ -7,3 +7,178 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+
+# Performances
+
+We compare the performances achieved with this version of the Cholesky factorization with the MKL one, and we obtain the next figure. Here the blocks are of size (256, 256).
+
+![Perf](scalability_lfaria.png)
```
ffevotte: Has this image has been obtained with the simplified code above? If it comes from the complete high-performance implementation in `TiledFactorization`, we should say so here.

ffevotte (commented):
Thanks for this write-up!

Here are my comments on the brand-new README.

ffevotte:
A global comment on this branch is: is there a reason why we need to have generated documentation files (such as `build/*.html`) checked in the git repo ?

README.md
maltezfaria: @ffevotte I agree with you. I wanted to mention this earlier, but forgot: we should discourage the use of `DataFlowTasks.sync`. In fact, perhaps we should be even more drastic and remove it altogether from the public API. As François mentioned, there are often better ways to sync the output before returning which will allow for better composability. The code in `TiledFactorization` already uses and identical approach https://github.com/maltezfaria/TiledFactorization/blob/a8a4f8e9c1cd347a6eaef60b0f38477077121908/src/cholesky.jl#L43-L44

vsivadon: I see, i just removed it at one point so it would make the profiling cleaner (because at the end there's juste one "Default" task).
I think we can keep it though, it's like the MPI.Barrier() function, not recommended but still useful sometimes, for debugging, easier developpement, safety etc...

README.md
```diff
@@ -7,3 +7,216 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+write!(X, alpha) = (X .= alpha)
+readwrite!(X) = (X .+= 2)
+n = 1000
+A = ones(n)
+function ex()
+    @dspawn readwrite!(@RW(A))              # 1
+    @dspawn write!(@W(@view A[1:500]), 2)   # 2
+    @dspawn write!(@W(@view A[501:n]), 3)   # 3
+    @dspawn readwrite!(@RW(A))              # 4
+end
```
ffevotte: either explicitly run `ex()`, or replace it with a `let` block (or a `begin` block)

README.md
ffevotte: Actually, I was thinking of renaming the serial version `cholesky_tiled!`, and keeping the name `cholesky_dft!` for the parallel version. This way we could have both versions alongside one another, which makes it easier for potential users to compare them. What do you think?

README.md
```diff
@@ -7,3 +7,216 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+            for tk in tj:tn
+                rj = tilerange(tj, ts)  ;  rk = tilerange(tk, ts)
+                mul!(view(A,rj,rk), adjoint(view(A,ri,rj)), view(A,ri,rk))
+            end
+        end
+    end
+
+    # Construct the factorized object
+    return Cholesky(A,'U',zero(LinearAlgebra.BlasInt))
+end
```
ffevotte: I think this version could benefit from the rewrite with `Atile` as well.

README.md
```diff
@@ -7,3 +7,216 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+            @dspawn ldiv!(@R(L), @RW(Atile[ti,tj])) label="ldiv ($ti,$tj)"
+        end
+
+        # Submatrix update (III)
+        for tj in ti+1:tn
+            for tk in tj:tn
+                Ajk = Atile[tj,tk]
+                Aik = Atile[ti,tk]
+                Aji = adjoint(Atile[ti,tj])
+                @dspawn mul!(@RW(Ajk), @R(Aji), @R(Aik), -1, 1) label="schur ($tj,$tk)"
```
ffevotte: I think we determined that it was better to express the read dependency to `Aij` directly instead of `adjoint(Aij)`.

README.md
```diff
@@ -7,3 +7,216 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+end
+```
+
+If you run this code, you'll see `memory_overlap` warning on the fact that there's no method to evaluate the memory overlaping between an `Adjoint` and a `SubArray`. So we have to implement this method :
+
+```julia
+using DataFlowTasks: memory_overlap
+memory_overlap(U::Adjoint,A) = memory_overlap(U.parent,A)
+memory_overlap(U,L::Adjoint) = memory_overlap(L,U)
+```
```
ffevotte: Please don't advise users to do this: it is type piracy

Well, technically it isn't here because this code defines a new `memory_overlap` function instead of adding methods to `DataFlowTasks.memory_overlap`, but this version shouldn't work either...

README.md
```diff
@@ -7,3 +7,216 @@ Status](https://github.com/maltezfaria/DataFlowTasks.jl/workflows/CI/badge.svg)]
...
+
+The next results are obtained with a slightly modified version of the function presented above. We used the `LoopVectorization` for the serial cholesky (I), `TriangularSolve`'s `ldiv!` and `Octavian`'s `matmul_serial!`.
+
+The package `TiledFactorization` contains this implementation, you can use it with the following lines :
+
+```julia
+import Pkg
+Pkg.add("https://github.com/maltezfaria/TiledFactorization.git")
+using TiledFactorization: cholesky!
+```
```
ffevotte: I think profiling should be illustrated on the example developed above. After all, we're not showing off with high-performance here, we are merely presenting a use-case.

IMO, only the performance plots should be coming from the `TiledFactorization` package.

README.md
vsivadon: Oh i understand, it's clearer this way.

ffevotte:
Commit 9085534 introduces a way to automatically generate the `README.md` file from a `Literate.jl` julia script (located in `docs/readme/README.jl`). A collateral benefit is that the same contents are now also generated in the form of a Jupyter notebook (which can be seen directly from [github](https://github.com/maltezfaria/DataFlowTasks.jl/blob/documentation/docs/readme/README.ipynb) or from [nbviewer](https://nbviewer.org/github/maltezfaria/DataFlowTasks.jl/blob/documentation/docs/readme/README.ipynb)).

The new version of `README.md` has been generated this way, and we can now be sure that all code in it is runnable (which incidentally increases the number of test cases, although these are run during documentation generation and are therefore not accounted for in the coverage analysis).

I'll now go ahead and try to resolve all remaining issues with `README.md`, which should get us close to the final contents. Maybe a last review from y'all would be good though, in order to make sure we're on the same page in terms of how we want to present `DataFlowTasks` to new users.

More work is still needed to finalize the full Documenter docs, especially if we want to include all examples. But maybe we can reach pretty quickly a "presentable" state, where all features are documented. And from there we can add more tutorials/examples (maybe using `DemoCards.jl` or `Literate.jl` to more easily ensure they are runnable).

docs/readme/README.jl
```diff
@@ -0,0 +1,247 @@
...
+# - read-write: `@RW` or `@READWRITE`
+#
+# Anywhere in the task body, a `@R(A)` annotation for example implies that `A` will be accessed in read-only mode by the task.
+
+#-
+
+# Let's look at a simple example:
+
+using DataFlowTasks
+DataFlowTasks.reset!() #hide
```
ffevotte, line 61: Unresolved questions regarding `reset`:
- do we want to keep it? (cf. #24)
- even if so, do we want to advertize its use?

OTOH: if we don't want to have/advertize it, make sure removing it doesn't affect the outcome of the examples shown here.

docs/readme/README.jl
```diff
@@ -0,0 +1,247 @@
...
+
+# (Also note how extra annotations were added in the code, in order to attach meaningful labels to the tasks. These will later be useful to interpret the output of debugging & profiling tools.)
+
+#-
+
+# The code below shows how to use this `cholesky_tiled!` function, as well as how to profile the program and get information about how tasks were scheduled:
+
+## DataFlowTasks environnement setup
+DataFlowTasks.reset!()
+DataFlowTasks.enable_log()
```
ffevotte, line 180: Are these lines needed? Would `resetlogger!` be enough?

docs/readme/README.jl
```diff
@@ -0,0 +1,247 @@
...
+A = A + n*I;
+
+#-
+@info "Testing sequential Cholesky factorization"          #hide
+F = cholesky_tiled!(copy(A), ts)                           #hide
+                                                           #hide
+## Check results                                           #hide
+err = norm(F.L*F.U-A,Inf)/max(norm(A),norm(F.L*F.U))       #hide
+@show err                                                  #hide
+@assert err < eps(Float64)                                 #hide
```
ffevotte, line 196: Naive question: why not something simpler, like
```
err = norm(F.L*F.U - A) / norm(A)
```
Also: I arbitrarily put the tolerance to machine precision, but this might not be the right thing to do.

ffevotte (commented):
I think the README is now in a state where it can be reviewed by anyone willing to.

I've commented on the few issues that still need to be discussed collectively IMO.

docs/readme/README.jl
ffevotte: Oh, and if we want to keep `reset()`, we'll have to add tests for it.

PR #26, opened 2022-08-18
Optional dependencies management

ffevotte:
This is a proposal for a tool allowing to conveniently load optional dependencies.

For example:
```julia
using DataFlowTasks
DataFlowTasks.@using_opt GLMakie
```

It currently works by temporarily activating a project located in `DataFlowTasks/optional-deps/Project.toml`, loading the requested dependencies, and finally re-activating the previous project. An other option could have been to fiddle with `LOAD_PATH`; not sure which idea is the cleanest...

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.1.0"
 [deps]
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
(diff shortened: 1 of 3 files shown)


maltezfaria:
I like this proposal. It is way cleaner than what I usually do as a user, which is to add these *optional dependencies* to the main Julia environment. I went over the implementation and it LGTM if you want to go ahead and merge it.

ffevotte:
Cool, thanks!

The only thing that worries me with the proposed implementation is its ability to age well when the system of the user will evolve around it: what happens when the julia version changes? Do we need to provide ways to update the optional dependencies themselves?

I think I want to make a few more tests before I merge this, but I'll do it quickly anyway.

ffevotte:
The release of Julia 1.8 gave me the opportunity to "battle-test" this feature in a situation where the Julia version changes: things went smoothly.

I'm sure there will come a time when the list of optional dependencies will need to be updated, or when the environment will need to get updated or resolved again. But I'm not sure how to handle such cases, and (who knows?) by then Julia might have support for proper sub-projects...

Anyways, I'm a bit more confident that this is useful, so I'm removing the "WIP" status of this PR.

PR #31, opened 2022-09-07
support for `v1.6`. Fixes #29

### LINKED ISSUE #29: Add support with `v1.6`
Currently, the package does not work with `v1.6` because of the `errormonitor` function.

maltezfaria:
Changes the `errormonitor` to `maybe_errormonitor`, which is a no-op if `VERSION < 1.7`.

One downside of this simple fix is that it seems we no longer properly handle errors in `v1.6`, and so some tests only pass on `>=v1.7`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,7 +10,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.7'
+          - '1.6' # LTS
+          - '1'
         os:
           - ubuntu-latest
         arch:
--- a/src/DataFlowTasks.jl
+++ b/src/DataFlowTasks.jl
@@ -1,5 +1,5 @@
 """
-    moduel DataFlowTask
+    module DataFlowTask
 
 Create `Task`s wich keep track of how data flows through it.
 """
@@ -10,6 +10,7 @@ const PROJECT_ROOT =  pkgdir(DataFlowTasks)
 using ThreadPools
 using DataStructures
 using Requires
+using Compat
 import Pkg
 
 """
--- a/src/DataFlowTasks.jl
+++ b/src/DataFlowTasks.jl
@@ -35,7 +35,7 @@ include("dataflowtask.jl")
 include("arrayinterface.jl")
 include("dag.jl")
 include("scheduler.jl")
-include("otherschedulers.jl")
+# include("otherschedulers.jl")
 
 export
     @dtask,
--- a/test/dataflowtask_test.jl
+++ b/test/dataflowtask_test.jl
@@ -100,13 +100,15 @@ end
     @test_throws TaskFailedException wait(t)
     @test isempty(ERRORS)
 
-    # Errors are handled when debug mode is on
+    # Errors are handled when debug mode is on.
     DataFlowTasks.enable_debug(true)
-    empty!(ERRORS)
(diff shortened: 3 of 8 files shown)


test/dataflowtask_test.jl
```diff
@@ -100,13 +100,15 @@ end
...
-    @test_throws TaskFailedException wait(t)
-    @test length(ERRORS) == 1
-    @test ERRORS[1][1].exception.msg == "Expected error"
+    if VERSION >= v"1.7" # FIXME
+        empty!(ERRORS)
+        t = @dspawn error("Expected error")
+        @test_throws TaskFailedException wait(t)
+        @test length(ERRORS) == 1
+        @test ERRORS[1][1].exception.msg == "Expected error"
+    end
```
maltezfaria: @ffevotte is it difficult to add support for the handling of errors in `v1.6`?

.github/workflows/CI.yml
```diff
@@ -1,3 +1,5 @@
+env:
+  JULIA_NUM_THREADS: 4
```
maltezfaria: @ffevotte With this we can run the CI using multiple threads. Assuming the CI machine has multiple cores, perhaps the figures be directly included from the CI for all but the `README.md` pages in the documentation?

src/DataFlowTasks.jl
```diff
@@ -35,7 +35,7 @@ include("dataflowtask.jl")
 include("arrayinterface.jl")
 include("dag.jl")
 include("scheduler.jl")
-include("otherschedulers.jl")
+# include("otherschedulers.jl")
```
maltezfaria, line 39: I am ignoring the `otherschedulers.jl` file for now, as well as its tests. If we decide that we have no plan to support other scheduler types, then lots of the code in the package can be significantly simplified. Something to think about...

.github/workflows/CI.yml
maltezfaria: Ok, so we better make sure that there are enough cores for the threads, or some of the tests will fail. It seems that the GitHub runners have two cores, so I am changing this to 2 threads. 

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

test/dataflowtask_test.jl
ffevotte: My feeling is that we don't really need `errormonitor` at all, and errors could be handled by  [`handle_errors`](https://github.com/maltezfaria/DataFlowTasks.jl/blob/main/src/utils.jl#L50-L62) alone. If that is the case, then test failures with 1.6 are not really related to `errormonitor`, but maybe something else.

But all these are only uneducated guesses at this stage. Let me try and reproduce errors locally; I'll investigate and report back.

ffevotte:
In the latest commit, I removed all uses of `errormonitor` which I think are not really necessary (do you agree @maltezfaria?)

A compatibility issues with v1.6 was caused by `DataFlowTasks.handle_errors` relying on the existence of `current_exceptions`, which was added in v1.7. Luckily, this feature is available via [`Compat.jl`](https://github.com/JuliaLang/Compat.jl).

Hopefully, all tests should now pass both on 1.6.5 and 1.8.0; there should be no need to conditionally disable some of them. And hopefully the user experience should be the same for all versions.

Issue #14, opened 2022-06-27
Simplify and improve critical path algorithm


maltezfaria:
We should avoid the creation of a dense adjacency matrix for doing this computation. Also, I think we could do something simpler than Djisktra here (see e.g. https://www.geeksforgeeks.org/find-longest-path-directed-acyclic-graph/)

_Originally posted by @maltezfaria in https://github.com/maltezfaria/DataFlowTasks.jl/pull/10#discussion_r907534675_

maltezfaria:
This was closed too early. Turns out the *critical path* algorithm is still quite brittle. It assumes, for instance, that the tags for the tasks start at `1`, if that is not the case it errors with an out-of-bounds access. Probably a rewrite of the logger and graph algorithms is in order.

ffevotte:
Agreed!
This might not be as straightforward as I initially thought, though: for example, how to know which task is the first one we want to log? Or even: is there necessarily a unique first task?

PR #34, opened 2022-09-08
Changes prior to first release

### LINKED ISSUE #30: Explain figures on `README` which are not generated by the `CI`
There is currently one figure for the scaling and comparison with MKL that lacks explanation on how (script) and where (machine) it was generated. A possible fix is to add a link to the script used to generate the figure on `TiledFactorization`.

maltezfaria:


src/logger.jl
```diff
@@ -100,42 +105,58 @@ Global `Logger` being used to record the events. Can be changed using [`setlogge
...
+    # check that logger contains tasks with contiguous tags (not necessarily
+    # starting at 1)
+    tasklogs = topological_sort(logger)
+    nbnodes  = length(tasklogs)
+    for i in 2:nbnodes
+        tag(tasklogs[i]) == tag(tasklogs[i-1]) + 1 || error("tasks not contiguous")
+    end
+    # HACK: because the current algorithm assumes the tasks start at one, apply
+    # a shift here and at the end
+    shift = tag(first(tasklogs)) - 1
```
maltezfaria, line 165: This is a hacky way to account for the fact that the critical path algorithm makes some strong (implicit) assumptions about the tags associated with the tags. In particular, it assumes that the tags start at one, which in turn means that the `resetlogger!` had to reset the task counter. Reseting the task counter should be avoided as it can mess up with the precedence notion of the tasks.

This is not really a fix for #14, but rather a quick-and-dirty workaround. I still think the graph algorithms should be made more robust in the future.

src/utils.jl
```diff
@@ -42,6 +42,24 @@ end
...
+`enable_graphics` for the first time may take a while.
+
+Available backends are: `CairoMakie` and `GLMakie`.
+"""
+function enable_graphics(;backend=:GLMakie)
+    supported = (:GLMakie, :CairoMakie)
+    backend in supported || error("supported backends: $supported")
+    @eval Main (DataFlowTasks.@using_opt GraphViz, $backend)
+end
+
```
maltezfaria: @ffevotte What do you think about a function to simplify the loading of the optional dependencies? This is in line with the other "enablers" we have (e.g. `enable_log` and `enable_debug`).

maltezfaria: This was a bad idea, so I went back to the original way you were doing things.

README.md
```diff
@@ -142,13 +169,13 @@ function cholesky_dft!(A, ts)
         # Left blocks update
         U = UpperTriangular(T[i,i])
         for j in i+1:n
-            @dspawn ldiv!(@R(U)', @RW(T[i,j])) label="ldiv ($i,$j)"
+            @dspawn ldiv!(@R(U'), @RW(T[i,j])) label="ldiv ($i,$j)"
```
ffevotte: I thought putting the data dependency on `U'` rather than `U` came with a performance issue... Is that not the case any more?

PR #37, opened 2022-09-19
Improve the critical path implementation

### LINKED ISSUE #14: Simplify and improve critical path algorithm

ffevotte:
This is still a WIP, but I'm opening the PR early to avoid duplicate work on this.

The idea here is to have a generic algorithm in `dag.jl`, which should be able to work on any graph that implements a few methods. The `logger.jl`, does not hold any critical-path-related code any more; it merely implements the required interface for the `Logger` type.

Existing tests should currently pass, but we should probably add a few more tests to check the correctness of the implementation in more "exotic" circumstances.


maltezfaria:
I went over the code and this exactly what we needed from the beginning -- thanks for handling it!

The only (minor) comment I have is that the `tag` method is also part of the required interface, and that it should never be `0` or the code will give the wrong answer.

src/dag.jl
```diff
@@ -276,40 +276,50 @@ end
...
+            pl, _ = lp[n]
+            if pl > path_length
+                path_length = pl
+                predecessor = n
             end
         end
+
+        # Augment the longest path with the weight of the current node
+        path_length += weight(node)
+        lp[tag(node)] = (path_length, predecessor)
```
maltezfaria, line 310: ~~Can't we just have the `lp` being a dict of the element type of the `graph` so that having a `tag` is not really needed?  The tag can always be used, it it exists, to hash the elements of the graph for better performance (currently the case for the `DataFlowTask`s).~~

Ok, I think I get it now. Everything is based on the `tag`s, and you need `inneighbors` to return the `tags` and not the nodes themselves. Sorry for the noise.

maltezfaria:
> 

I think I jumped into conclusion a bit too quickly before really understanding the implementation. I am a bit confused since sometimes you use the `tag`, and sometime the node itself...

It seems that you assume that `inneighbors` returns the `tags` of the nodes, and not the node themselves. Is that correct?

src/dag.jl
```diff
@@ -276,40 +276,49 @@ end
...
+    longest_path(graph) -> path
+
+Finds the critical path of a DAG `graph`, and returns the nodes constituting it
+in reverse order.
+
+Required interface for the graph:
+- topological_sort(graph) -> collection of nodes in `graph`, sorted in topological order
+- inneighbors(node)       -> collection of predecessors of `node` (identified by their tags)
+- weight(node)            -> cost of `node` (a positive Float64)
+- tag(node)               -> id of `node` (a strictly positive Int)
```
ffevotte: > the tag method is also part of the required interface, and that it should never be 0 or the code will give the wrong answer.

Precisely! I tried specifying the broad lines of the interface in this comment. Does this answer your questions?

Since `Logger` is currently the only type implementing the interface, I was lazy and tried to re-use as many existing methods as possible. For example: `topological_sort` and `tag` already existed. `inneighbors` also existed (not as a method but as a field), and it identifies predecessors by their tags.

I decided to re-use all that in order to minimize the impact of the critical path algorithm on the code in `logger.jl`. An other option would have been to design the desired interface independently, and to adapt the `Logger` implementation to conform to it. It is not too late to adopt such a strategy.

maltezfaria: > Precisely! I tried specifying the broad lines of the interface in this comment. Does this answer your questions?

👍

maltezfaria: > Since `Logger` is currently the only type implementing the interface, I was lazy and tried to re-use as many existing methods as possible. For example: `topological_sort` and `tag` already existed. `inneighbors` also existed (not as a method but as a field), and it identifies predecessors by their tags.

Makes sense. The thing that got me confused is that the `DAG{DataFlowTasks}` has an `inneighbors` method that returns the nodes and not the tags. Also, it requires you to pass the `dag` as well since `DataFlowTask` does not store the neighbours (unlike the `TaskLog`) in its fields. Perhaps we should rename `inneighbors` to something like `intags` for the interface method to avoid possible confusion? Otherwise I am quite happy with the changes!

maltezfaria: > I decided to re-use all that in order to minimize the impact of the critical path algorithm on the code in `logger.jl`. An other option would have been to design the desired interface independently, and to adapt the `Logger` implementation to conform to it. It is not too late to adopt such a strategy.

I think your strategy is simple and works for what we need. In the future, should we need to compute the longest path for scheduling purposes on the flight, we can always add the interface you described to the `TaskGraph` type. Requiring a strictly positive (integer) tag is not something every dag would have I guess, but we are not building a graph package right? 🤣

maltezfaria (changes requested):
Besides the suggestion to rename `inneighbors` to avoid confusion with the existing method (which returns something slightly different), this PR looks good to me and I suggest we go ahead and merge it. 

Thanks!

src/dag.jl
ffevotte: > Perhaps we should rename inneighbors to something like intags for the interface method to avoid possible confusion?

Nice idea. Done!

ffevotte: > In the future, should we need to compute the longest path for scheduling purposes on the flight, we can always add the interface you described to the TaskGraph type. Requiring a strictly positive (integer) tag is not something every dag would have I guess, but we are not building a graph package right?

Yes, we can always implement the interface for the `TaskGraph` type later. But if/when that time comes, we can also change the interface in order for it to be easier to implement for `TaskGraph`. And it will be no big deal to adapt the implementation for `Logger` as well.

I considered adding functions like `tagtype(node)` and maybe re-use `zero(tagtype(node))` as the sentinel value for root nodes. This would make everything more generic, but it feels like it would be a bit over-engineered at this time.

maltezfaria (approved):

maltezfaria:
Looks good to me. If we ever need an *on-the-flight* longest path calculation we can revisit the interface to make it work for `TaskGraph`s. Merging this now.

Issue #40, opened 2022-10-07
Document troubleshooting and debugging strategies


maltezfaria:
We should improve the documentation of how to debug `DataFlowTasks`. In particular, we should mention the `force_sequential` method as a first check. It turns out that a lot of the times the problem is with the serial code, and it is much easier to read the serial stack trace than the parallel one!

maltezfaria:
Make sure to also mention `restart_scheduler!` (speaking from personal experience here 😅 )

maltezfaria:
Ok, a few more hours implementing the hiearchical `LU` on `HMatrices.jl` using `DataFlowTasks`. It works now, but here are some lessons learned:
- *Debugging `DataFlowTasks` is hard*. The workflow I found most useful was:
    1. Write the serial code, make sure it work
    2. Add the `@dspawn` to the serial code
    3. Run the code with `force_sequential` and make sure it still works
    4. Run the code with a new `force_linear_dag()=true` (name to be discussed). This makes sure `DataFlowTask`s are created and schedule, but it forces the underlying `DAG` to be linear (i.e node `i` is always connected to node `i+1`. This is closer to the *real thing* than `force_sequential`, and was quite helpful as a debugging tool. One subtlety I encountered had to do with the behaviour of captured variables (#25), and step `iv` helps to debug this. 
    5. If `iv` works, then deactivate the `force_linear_dag()` and see if tests pass. If they don't, the problem is in the data dependencies. Look at the code, scratch your head, and continue debugging 😕 

- Things are tricky when a `DataFlowTask` creates data, and some subsequent tasks need to use it. For example:
```julia
t1 = @dspawn (sleep(1); ones(2))
t2 = @dspawn
    sum(fetch(t1))
end
```
Fetching `t2` will work, but the internal task graph does not know explicitly that `t2` depends on `t1`, so lots of the logging information obtained are misleading. In fact, it will log `t2` as taking roughly `1` second to run. Ideally, we would not start running the code in `t2` task until `t1` is finished, and for that there should be an edge `t1 --> t2`. Can we automatically add that in the example above? If not, how do we handle the case where **data is created** inside a task? One alternative is to do
```julia
t2 = @dspawn
    sum(fetch(@R(t1)))
end
```
which will add `t1` as a read dependency to `t2`. We then need to slightly change how dependencies are computed to treat the case where some of the `data` is itself a `DataFlowTask`. Seems a bit hacky...

[...]

PR #42, opened 2023-07-11
Transition from Requires.jl to v1.9-style weak dependencies

ffevotte:
Now that 1.9 allows for "weak dependencies", I wanted to test whether we could use these to handle optional dependencies to GraphViz & Makie for visualization.

This is (the beginning of) an attempt to do this. It's not finished (yet), but I wanted to discuss this as early as possible in order to make an informed decision about where we go with this.

One first thing to note is that it is nice to have these TTFX-heavy workloads inside precompileable modules. The drawback is that these modules are "insulated" from `DataFlowTasks`: the visualization code used to be able to use anything from `DataFlowTasks`, and also used to be able to define new functions inside the `DataFlowTasks` namespace. This is not the case anymore; the good practice seems to use extension code to add methods to existing functions, which requires our old Requires-based code to be updated in one of two ways (there might also be more ways to do things that I haven't thought about):

1. extend functions from the public API of `Makie`/`GraphViz`. For example, I replaced our custom `plot_traces` and `plot_dag` with new methods for `Makie.plot` and `GraphViz.Graph`. Below is an excerpt from the updated `README` (other examples are not up-to-date):
    ```julia
    log_info = DataFlowTasks.@log cholesky_dft!(A ,ts);
    dag = GraphViz.Graph(log_info)                               # formerly plot_dag(log_info)
    trace = plot(log_info; categories=["chol", "ldiv", "schur"]) # formerly plot_traces(log_info; categories)
    ```
2. extend functions from `DataFlowTasks` itself. This means that `DataFlowTasks` might define functions which are available but unusable at first because they have no method until an extension is loaded. I did this for example with the utility function `savedag`, which unfortunately does not have an equivalent in `GraphViz`'s public API.

Another thing to take care of is to make sure that unit tests can use functions internally defined in the extensions. This adds a few lines at the top of `loginfo_test.jl`, but it also provides a clean way to check whether extensions are actually loaded, so all in all I'm rather happy with this.

So I guess the next questions are:

1. do we want to go along this path and declare `GraphViz` and `Makie` as weak dependencies? The associated TODO list would IMO mostly involve adapting the documentation and examples to the new API (with a bit of a question mark related to how `Documenter` handles extensions)
2. do we want to keep support for `Requires`-based optional dependencies for users of julia v1.8 and below? It looks like this is feasible, but it would involve a few extra steps.
3. Are we happy with the accompanying API changes? It would be relatively easy to revert them and keep the original API, defining empty functions in `DataFlowTasks` that would not be usable until extensions are loaded.


maltezfaria:
I like the move towards `WeakDeps`, and the precompilation benefits that come with it! 

I think the "drawbacks", if you can call it that, of having to either extend methods of the loaded extension or of the *main* package, are nothing really major. I don't see any pain points there.

As for providing [backward compatibility](https://pkgdocs.julialang.org/v1/creating-packages/#Backwards-compatibility) with pre `1.9` Julia versions, I don't think we should care too much for now. We can always fix this later if needed.

ffevotte:
This should be more or less complete now. Since extension modules are not directly available, documenting the interface they introduce proved to be a bit more involved than expected, but all examples and docs should now be up-to-date with the new way of doing things.

One thing we'll have to do at some point before the next release: update `Project.toml` to
1. specify compat bounds for weak dependencies (I think this is mandatory), and
2. bump the version number (I guess to 0.2.0 to signal a breaking release, even though the breaking parts are in extensions).

maltezfaria:
LGTM (there is one lingering reference to Requires in the `README.jl`)

Edit: I went ahead and fixed it, as well as a few typos.

maltezfaria:
> One thing we'll have to do at some point before the next release: update `Project.toml` to
> 
> 1. specify compat bounds for weak dependencies (I think this is mandatory), and
> 2. bump the version number (I guess to 0.2.0 to signal a breaking release, even though the breaking parts are in extensions).

Should we discuss some other (minor) improvements for `0.2`? I mentioned some pain points in #40, and would be happy to put in the time to implement a few new features.

docs/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
+DataFlowTasks = "d1549cb6-e9f4-42f8-98cc-ffc8d067ff5b"
```
maltezfaria, line 4: I never add the project itself as a dependency to `docs`, as I was under the impression this was not necessary (`Documenter` looks into its parent folder for a project right?).

ffevotte: I don't think `Documenter` looks into its parent folder to find the documented project, so at the time when the documentation is built, the parent package has to be present in the documentation environment in some way. However, one of the steps in the GitHub CI action takes care of `Pkg.dev`ing the parent project in the documentation subproject. So both ways work: if the committed `docs/Project.toml` does not mention `DataFlowTasks`, things will get set up correctly just ahead of time.

I've recently seen packages where even the `docs/Manifest.toml` was committed so as to make sure that the parent project is `dev`ed from the parent directory. [`Graphs.jl`](https://github.com/JuliaGraphs/Graphs.jl/tree/master/docs) is an example of this.

Personally, I'd be in favor of keeping `DataFlowTasks` as a documentation dependency in the committed `docs/Project.toml`, because this simplifies the process to locally build the documentation. But I'm open to compromise :smile:

ffevotte:
> Should we discuss some other (minor) improvements for 0.2? I mentioned some pain points in https://github.com/maltezfaria/DataFlowTasks.jl/issues/40, and would be happy to put in the time to implement a few new features.

I've opened #43 to list the ideas that came to my mind when re-opening this project after a few months.

maltezfaria:
Can we go ahead and merge this?

ffevotte:
Yes, we can merge this.

The only unresolved issue is related to the docs dependencies, so nothing specific to the weakdeps transitions and we can decide later. I'm going to mention this in #43 so that we don't forget to decide about this.

PR #45, opened 2023-07-13
Improve tests with `Aqua.jl`, simplify `AbstractArray` interface, and update CI

maltezfaria:
- Fixes method ambiguities
- Remove stale dep on `StaticArrays`
- Add`Aqua` badge
- Simplify the array interface using the `parent` method
- Update CI to best practices (triggers only on main, tags, and PRs)


maltezfaria:
This is ready for review. I encountered two (minor) issues:

- `Aqua.jl` has a test to check the format of the `.toml` files. Because of the `[weak deps]` fields, that test fails on `1.6`. I did not find an easy way to circumvent that in the `Aqua` documentation, so I simply skipped the `toml` format tests on versions `<1.9`. 
- Fixing the method ambiguities introduced lines that were not covered by the tests, triggering a failure of the `codecov` action. Can we be more loose on this check?

ffevotte:
> Aqua.jl has a test to check the format of the .toml files. Because of the [weak deps] fields, that test fails on 1.6. I did not find an easy way to circumvent that in the Aqua documentation, so I simply skipped the toml format tests on versions <=1.9.

It looks to me like this was the correct way to handle things

> Fixing the method ambiguities introduced lines that were not covered by the tests, triggering a failure of the codecov action. Can we be more loose on this check?

I think we could change the default Codecov settings using [codecov.yml](https://github.com/maltezfaria/DataFlowTasks.jl/blob/main/codecov.yml). But we could also simply keep those as a safeguard, and choose to ignore the failures in cases like this.

Regarding the ambiguities themselves, it seems to me that we're hitting here an inconvenience related to double dispatch on the arguments of `memory_overlap`: if there were more types to special case, the number of combinations needed to disambiguate would explode. But this is only because we handle this via multiple dispatch on both arguments of `memory_overlap`, when in reality what we want in this specific case is a form of single dispatch: for some types like `TriangularArray` or `Adjoint`, we don't want to directly check whether the instance overlaps with something else, we want to get the real data behind a given instance and check that instead.

So maybe we could devise an 2-stage implementation that would scale better in this respect, and avoid a lot of ambiguities: instead of checking the overlap between two pieces of data `di` and `dj` using [`memory_overlap(di, dj)`](https://github.com/maltezfaria/DataFlowTasks.jl/blob/main/src/dataflowtask.jl#L126), we could use something like:
```julia
memory_overlap(memory_data(di), memory_data(dj))
```
where we have defined
```julia
memory_data(d) = d  # By default this return the data itself

# one method for each special cased type
memory_data(t::LinearAlgebra.AbstractTriangular) = t.data
memory_data(a::Adjoint) = t.parent
```

It seems to me that this would work, and avoid the combinatorial explosion we're begininng to see here. It would however slightly complicate the interface for memory overlap detection: there are now two complementary mechanisms which need to be documented.

What do you think?

maltezfaria:
> > Aqua.jl has a test to check the format of the .toml files. Because of the [weak deps] fields, that test fails on 1.6. I did not find an easy way to circumvent that in the Aqua documentation, so I simply skipped the toml format tests on versions <=1.9.
> 
> It looks to me like this was the correct way to handle things
> 
> > Fixing the method ambiguities introduced lines that were not covered by the tests, triggering a failure of the codecov action. Can we be more loose on this check?
> 
> I think we could change the default Codecov settings using [codecov.yml](https://github.com/maltezfaria/DataFlowTasks.jl/blob/main/codecov.yml). But we could also simply keep those as a safeguard, and choose to ignore the failures in cases like this.
> 

👍 

> Regarding the ambiguities themselves, it seems to me that we're hitting here an inconvenience related to double dispatch on the arguments of `memory_overlap`: if there were more types to special case, the number of combinations needed to disambiguate would explode. But this is only because we handle this via multiple dispatch on both arguments of `memory_overlap`, when in reality what we want in this specific case is a form of single dispatch: for some types like `TriangularArray` or `Adjoint`, we don't want to directly check whether the instance overlaps with something else, we want to get the real data behind a given instance and check that instead.
> 
> So maybe we could devise an 2-stage implementation that would scale better in this respect, and avoid a lot of ambiguities: instead of checking the overlap between two pieces of data `di` and `dj` using [`memory_overlap(di, dj)`](https://github.com/maltezfaria/DataFlowTasks.jl/blob/main/src/dataflowtask.jl#L126), we could use something like:
> 
> ```julia
> memory_overlap(memory_data(di), memory_data(dj))
> ```
> 
> where we have defined
> 
> ```julia
> memory_data(d) = d  # By default this return the data itself
> 
> # one method for each special cased type
> memory_data(t::LinearAlgebra.AbstractTriangular) = t.data
> memory_data(a::Adjoint) = t.parent
> ```
> 
> It seems to me that this would work, and avoid the combinatorial explosion we're begininng to see here. It would however slightly complicate the interface for memory overlap detection: there are now two complementary mechanisms which need to be documented.
> 
> What do you think?

I like your idea of having an intermediate step. As it turns out, the `parent` method of an `AbstractArray` does precisely what we want! 

We provide the `AbstractArray` interface to showcase the package's utilities, but it could as well live outside of `DataFlowTasks`, so it seems ok to leave the responsibility of properly implementing the `memory_overlap` method to the user. That way we don't have to document some low-level implementation detail that does not concern the package itself.

ffevotte:
This looks really good! I like how the `memory_overlap` is simplified by this change. (And it has the added benefit that the documentation does not need to be expanded)

Since this PR is all about CI & tests, I wanted to take the opportunity to change the way CI is triggered on github. I updated `CI.yml` to the current practice promoted by `PkgTemplates`; with this new setup CI should be triggered only once when we push to a PR branch. But it should still happen when we push to `main` directly.

Feel free to merge the PR if you like this change (or revert it if you don't like it)

PR #46, opened 2023-07-13
New implementation of weak dependencies handling

### LINKED ISSUE #47: Make sure @using_opt works as expected
I'm not sure `@using_opt` is as robust as we'd like, especially:
  - in cases where `DataFlowTasks` is used as a regular dependency in the current environment (as opposed to being the active project itself), and
  - when the Julia version changes.

ffevotte:
This is an entirely new implementation of weak dependencies handling, which replaces `@using_opt`. The key aspects of the new implementation are:

- the environment used to provides weak dependencies is created in a ["scratch space"](https://github.com/JuliaPackaging/Scratch.jl)
  - we can (and do) have one such environment per minor Julia version => more robust when switching versions
  - it is relocatable
- the weak dependencies environment is pushed to the Environment stack (as opposed to the old implementation which temporarily activated it while loading the required package)
  - this seems to be somewhat more robust when handling extensions (as it mirrors one of the intended use cases where the weak dependency can be found in the global `@vX.Y` env)

The new way of using it is therefore much simpler:
```julia
using DataFlowTasks
DataFlowTasks.stack_weakdeps_env!() # call this once to set up the environment stack
using GraphViz, GLMakie
```

The relevant environment is retrieved from the package scratch space, created if it did not exist, and instantiated if needed. These last 2 steps can be time consuming, especially the first time a weak dep is used in a given context. Using `stack_weakdeps_env!(verbose=true)` can help convince oneself that something is actually happening in the background.

Overall, I'm relatively happy with this, to the point that `test` and `doc` subprojects don't depend on `GraphViz` and `CairoMakie` anymore; instead they make use of `stack_weakdeps_env!()` to load optional dependencies when needed. Hopefully this won't have much impact on CI run times, but we'll see...


maltezfaria:
This is simpler and shorter than the previous implementation, so I think we can go ahead with it. A rebase is needed though now that we have merged #45 into main.

maltezfaria:
> This is simpler and shorter than the previous implementation, so I think we can go ahead with it. A rebase is needed though now that we have merged #45 into main.

But to be completely honest, I don't think I understand the main use case for this 😅 

We are trying to help users set up their environment so that they can use the optional deps in a *clean* way, but should that be part of the API? Do other packages propose similar methods?

ffevotte:
> We are trying to help users set up their environment so that they can use the optional deps in a _clean_ way,

Yep, that's the main use case (knowing that users include CI). A nice benefit of loading optional dependencies from a separate environment that is only instantiated when needed , is that CI tests take less than 2 minutes for v1.6.

> Do other packages propose similar methods?

Not that I know of. But AFAIU extensions are mostly used to provide integration between 2 packages that the user decides to use. In such cases, both packages have a good reason to be included in the environment. Here, we use extensions to provide additional features related only to debugging, so that there is not really any logical environment where these optional deps would live.

<br/>

On the other hand, you're right in that it does look like we're hitting a lack of tooling around stacked environments and context-dependent dependencies that is more general than a `DataFlowTasks`. See for exemple this very recent discourse thread: [Shortcut to load a package from another environment](https://discourse.julialang.org/t/shortcut-to-load-a-package-from-another-environment/101620).

So another point of view could be that such tooling should available globally, and DataFlowTasks shouldn't worry about it too much...But then wouldn't we at least need to document how to do things?

maltezfaria:
I see the advantages of `stack_weakdeps_env!` now, thanks. I still wonder whether this should be achieved by external tools or not. But while such tools don't exist, let's keep it! I will go over this PR and merge it if you are ok with that. 

p.s. The problem was mostly my lack of understanding, sorry

maltezfaria:
> Not that I know of. But AFAIU extensions are mostly used to provide integration between 2 packages that the user decides to use. In such cases, both packages have a good reason to be included in the environment. Here, we use extensions to provide additional features related only to debugging, so that there is not really any logical environment where these optional deps would live.

This is a good point too. Ok, I am convinced about `stack_weakdeps_env!`. If anything, it simplifies our own CI setup.

.github/workflows/CI.yml
```diff
@@ -21,6 +21,7 @@ jobs:
           - x64
     steps:
       - uses: actions/checkout@v2
+      - run:  perl -ni -e 'print unless /GLMakie/' ext/Project.toml
```
maltezfaria, line 24: Why do we need this?

maltezfaria (commented):
There was only one line in the docs CI that I didn't understand. The rest all looks good to me.

Thanks for putting this together, I am starting to like this `stack_weakdeps_env!` approach!

.github/workflows/CI.yml
ffevotte: This removes `GLMakie` from the weak dependencies environment. It's just a sort of optimization: since it is the heaviest package we have and we don't need it in CI, this saves some time while instantiating the environment.

PR #52, opened 2023-07-19
Changes to API

### LINKED ISSUE #49: API change / simplify internals of `DataFlowTasks`?
I have started to wonder if we need the `DataFlowTask` type at all. It would be nice if all we did was run some pre- and post-hook when a `Task` is created in order to keep track of data info (possibly in a global variable) and add some `wait`s depending on data dependencies. We will most likely still need a DAG-like to compute dependencies, but that is easy.

This would simplify the package's logic, and make it very clear that the sole role played by `DataFlowTask` is to automatically add some synchronization between tasks based on data. I am not sure whether this is possible to do in practice. Can we, for instance, simply overload the `Task` constructor?

```julia
function Task(func;data,mode)
        # compute dependencies using some global DAG
        deps  = compute_deps(data,mode) 
        # create the task
        tj = @task do
           # append dependencies
           for ti in deps
              wait(ti)
           end
           # run user's code
           res = func() 
           # post-hook, move itself to a finished Channel so that it can be removed from the global DAG
           add_to_finished!(tj)
           res
        end
        # add the task to the global DAG
        add_to_dag!(tj) 
        return tj
end
```

We would then need to adapt the macros accordingly, but if that is possible we would have a syntax like:

```julia
using DataFlowTasks
A = Vector{Float64}(undef, 4)
result = let
    @spawn fill!(@W(A), 0)           # task 1: accesses everything
    @spawn @RW(view(A, 1:2)) .+= 2   # task 2: modifies the first half
    @spawn @RW(view(A, 3:4)) .+= 3   # task 3: modifies the second half
    @spawn @R(A)                     # task 4: get the result
end
fetch(result)
```

Where we create `Task` objects, and not `DataFlowTask` objects.

Any thoughts?

maltezfaria:
The main usage is now through the `DataFlowTasks.@spawn` macro, which works very much like `Threads.@spawn`. In particular, it now returns a `Task` and not a `DataFlowTask`.

IMO this makes understanding and using the package easier. 

Left to do:
- [x] Update `TiledFactorization` to this new API
- [x] Check for performance regressions
- [ ] Go over the docs gain. I made some major changes that need to be checked.


maltezfaria:
As per 5da6c9a, there are no significant performance changes in the cholesky factorization compared to previous results (it does require the fix 831aa5d136d5a5b1fbcddc6acf366c346f438371, which has not yet been merged into main).

maltezfaria:
@ffevotte I think this is ready to be looked at (no pressure). The changes are conceptually simple, but a lot of code was modified in the process. 

Maybe we could also merge it first (all tests+docs are passing) and field-test it while implementing the other features? I am OK either way.

JieRen98:
Please merge it, thx!

JieRen98:
Congrats!

maltezfaria:
I think this was mostly ready, but still requires some field testing. Let me know if you run into any issues!

PR #55, opened 2023-10-20
Generate examples notebook in the `CI` process

maltezfaria:


maltezfaria:
Temporary docs can be previewed [here](https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR55/).

maltezfaria:
@ffevotte : for the examples generated using `Literate`, I think we can (safely) switch to the proposed workflow where the `.md` and `.ipynb` are generated during the `CI` on a self-hosted machine. This increases reproducibility, yields faster build times on the CI, and gives us the ability to showcase larger examples. It is also, IMO, simpler than the previous manual "generate locally and push" approach. 

For the `README`, I am on the fence on how to proceed. I see a few options:

- We change to having a "classic" `README.md` file on the root directory, with the required images pre-generated. This means it is our responsibility to make sure it is up to date.
-  We change to having a "classic" `README.md` file on the root directory, but we use images that are generated with the `CI`. That means we need to add links the corresponding `github.io` image. 
- We build the `README.md` from the `README.jl` locally, and then push. Possibly with a check on the `CI` that the `README` is up-to-date. This is what we have been doing so far, and while the more robust method, it does require a bit of boilerplate code (and manual steps) to keep things running smoothly (I am thinking about all those commits to update the examples/readme 😁 )

Any preferences or suggestions?

ffevotte:
> For the `README`, I am on the fence on how to proceed.

Yes, I do understand your concerns. On the other hand, I like the idea that code presented in the README is always valid (a bit like doctests): since it is one of the first things potential users are going to see, it is IMO good if we can ensure that they always work.

Maybe we could keep the idea of having the README.md coming from a README.jl, but try and reduce the machinery to make this working, and push as much as we can to the CI. I think an objective could be for things to work mostly like doctests: CI makes sure they are runnable and detects when the output is not up-to-date.

Implementing this should be relatively straightforward, the main difference being that CI would only check that the commited `README.md` contents are consistent with the automated run (whereas we currently have a system based on source code hashes -- not markdown contents -- which is complicated and overly conservative). Let me try and see whether I can suggest a simplified way to handle the README.

ffevotte:
Hey,

I think I have some ideas about how to proceed with the README. How would you like to proceed? Maybe merge this PR first, then I can base a new branch off of it to discuss the details related to the README. Would that be OK with you?

PR #57, opened 2023-10-23
Simplify README.md generation

ffevotte:
This PR (tries to) simplifies the README.md generation process:

- all files are now under `docs/src/readme` in order for everything to be deployed in the docs site
- the "source of truth" is still the README.jl literate source file
    - during CI (the following is done by `docs/Make.jl`):
        1. the script itself is run: this is needed to generate the included pictures, and also makes sure that everything works
        2. its notebook form is generated
        3. its markdown form is generated and links to pictures are updated to point at the documentation website
        4. a warning is printed if the contents of the generated `docs/src/readme/README.md` do not match the committed `/README.md` (and an error is thrown if this happens during CI)
     - during development:
         - in most cases nothing is needed: the commands advertised in the README keep working, and we don't care much about either the textual output (not included in the markdown document) nor the images (they will get updated as soon as the documentation is put on line)
         - when the README contents themselves are modified, we simply need to locally build the documentation, we'll get a warning about the README.md getting out of date, and we only need to
                 ```
                 cp docs/src/readme/README.md README.md
                 ```

Hopefully, links in the `README.md` point to correct URLs, but they will be broken at least until we push this PR and a new version of the documentation (including README images) gets depolyed.


ffevotte:
This should be ready for review. (But we won't know for sure that all links are correct until the docs are deployed "for real" at least once)

In any case, the notebooks and images generated during CI can be previewed here:
https://github.com/maltezfaria/DataFlowTasks.jl/tree/gh-pages/previews/PR57/readme

docs/make.jl
```diff
@@ -4,18 +4,42 @@ using Literate
...
+
+    # Generate notebook
+    Literate.notebook(src, pwd())
+
+    # Generate markdown
+    # -> fix image paths to link to github.io
+    Literate.markdown(src, pwd(), flavor=Literate.CommonMarkFlavor())
+    contents = read("README.md", String)
+    contents = replace(contents, "![](" => "![](https://maltezfaria.github.io/DataFlowTasks.jl/dev/readme/")
+    write(joinpath("README.md"), contents)
```
maltezfaria: Why is `joinpath` needed?

ffevotte: `joinpath` isn't useful. But reviews are!

maltezfaria:
LGTM. I think the `CI.yml` can be cleaned up a bit (my fault). Can I push here?

ffevotte:
> Can I push here?

Please do!

In the meantime, I'm working on a new Longest Common Subsequence example

maltezfaria:
Is this a bug?

https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/6614873153/job/17965773614?pr=57

I can't reproduce it locally, so I am very puzzled as to what may have gone wrong 🤔

maltezfaria:
> Is this a bug?
> 
> https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/6614873153/job/17965773614?pr=57
> 
> I can't reproduce it locally, so I am very puzzled as to what may have gone wrong 🤔

Re-running the CI on the same commit to reproduce the problem, and it is gone:

https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/6614873153/job/17966604038

This may very well be one of those elusive bugs... I will keep looking

maltezfaria:
Instead of polluting this `PR` with attempts to fix the (maybe?) bug in the `logger`, I will merge this and create a separate `PR` to investigate and discuss the issue.

maltezfaria:
But just for the record, the problem seems to appear every time the docs are generated on a *new build*, and it shows up again here:

https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/6615639951/job/17968211669?pr=57

FWIW, this may be a long and dormant bug, and the changes I made in how the documentation `CI` are triggered made it more frequent. I may have identified some possible issues with a race condition in the logging system, so I will look into that.

maltezfaria:
Ok, I checked the `README.md` after the merge, and the links appear to be correct!

And issue #59 discusses a possible data-race issue with the current logging system.

PR #58, opened 2023-10-23
New example: Longest Common Subsequence

ffevotte:
Preview:
https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR58/examples/lcs/lcs/


ffevotte:
This should be mostly ready for review.

Now that the documentation is generated on a powerful (and stable) system, I thought we could include some timings and discuss performances (for this new LCS example as well as the Blur-Roberts one). However:
- I don't observe the same behavior on the CI system as on my machine, and I'm not entirely sure that the performances are completely reproducible, even with this setup
- I had to deactivate code coverage when building the docs, otherwise performance measurements are completely off (especially for parallel computations)

So all-in-all, I could revert to a strategy similar to what we initially had with Blur-Roberts: not show any performance measurements and only discuss the profile picture produced by DataFlowTasks. @maltezfaria do you have any opinion on that?

maltezfaria:
I just triggered the `CI` twice on your last commit, and I obtained these two figures:

29b4a264

7aa5a311

It seems to me that the results are roughly the same, so it appears to me that the performance are *reproducible* in the strict sense that they give the same results twice on the same machine. 

Now, IIUC, you see something different on your local machine. Anything special about your machine (e.g. you are not on an M1/M2 right?).

I would be in favour of adding/keeping the discussion on the performance, but add the usual *caveats* that performance is really hard to measure correctly, and the precise conclusion or *tuning* are likely to be machine dependent. 

I will go over this example later today, thanks for setting it up!

ffevotte:
Yes, the other picture is bothering me more: on my machine I always see a performance improvement for the tiled version, due to cache effects. It may be relatively small, but is always there, so I thought it was not pure noise:
lcs-a30c7480

On the CI system, it looks like the sequential non-tiled and tiled versions are not always ranked in the same order.
But we could also adapt the text to explain that cache effects may or may not matter depending on the system (and I suspect that `BenchmarkTools` hides some of the effects by always working with a hot cache)

.github/workflows/CI.yml
```diff
@@ -63,9 +65,11 @@ jobs:
       - name: Install dependencies
         run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
       - name: Build and deploy
+        run: julia --project=docs/ docs/make.jl
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
-        run: julia --project=docs/ --code-coverage=user docs/make.jl
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run examples
+        run: julia --project=docs --code-coverage=user docs/run_examples.jl
```
maltezfaria, line 72: I think this is a good idea. My only concern is if it significantly increases the `CI` time on the Github hosted machines. Do you know if that is the case?

docs/run_examples.jl
```diff
@@ -0,0 +1,18 @@
+using Test
+
+@testset "Examples" verbose=true begin
+    examples_root = realpath(joinpath(@__DIR__, "../docs/src/examples"))
+    for name in readdir(examples_root)
+        dir = joinpath(examples_root, name)
+        script = joinpath(dir, "$name.jl")
+        ispath(script) || continue
```
maltezfaria: Isn't this a path by construction?

docs/src/examples/blur-roberts/blur-roberts.jl
```diff
@@ -50,20 +47,17 @@ end
...
+#
+# The `blur!` function averages the value of each pixel with the values of all
+# pixels less than `width` pixels away in manhattan distance. In order to
+# simplify the implementation, the filter is applied only to pixels that are
+# sufficiently far from the boundary to have all their neighbors correctly
+# defined.
+#
+# Results are written in-place in a pre-allocated `dest` array. Unless otherwise
+# specified, the filter is applied to the whole image, but can be reduced to a
+# tile if a smaller `range` argument is provided.
```
maltezfaria, line 60: Out of curiosity, why do you prefer this over the block-comment syntax?

docs/src/examples/blur-roberts/blur-roberts.jl
```diff
@@ -140,22 +119,22 @@ tmp  = similar(mat)
...
-
-#=
-
-## Tiled filter application
-
-The [`TiledIteration.jl`](https://github.com/JuliaArrays/TiledIteration.jl)
-package implements various tools allowing to define and iterate over disjoint
-tiles of a larger array. We'll use it to apply the filters tile by tile.
+# The elapsed time in this sequential version will serve as reference to
+# evaluate the performance of other implementations
```
maltezfaria: add `:` at the end

docs/src/examples/blur-roberts/blur-roberts.jl
```diff
@@ -140,22 +119,22 @@ tmp  = similar(mat)
...
-=#
+# ## Tiled filter application
+#
+# The [`TiledIteration.jl`](https://github.com/JuliaArrays/TiledIteration.jl)
+# package implements various tools allowing to define and iterate over disjoint
+# tiles of a larger array. We'll use it to apply the filters tile by tile.
+#
+# The `map_tiled!` higher-order function automates the application of a filter
+# `fun!` on all pixels of an image `src` decomposed with a tilesize `ts`. This
+# higher-order function is then used to defined tiled versions of the blur and
```
maltezfaria: defined -> define a

docs/src/examples/lcs/lcs.jl
maltezfaria, line 1: I need some time to read and digest this one, as I am unfamiliar with the algorithm. A general remark though: why did you not use a `setup` stage in the benchmarks here? We probably want to avoid the *hot cache* bias right?

maltezfaria:
> Yes, the other picture is bothering me more: on my machine I always see a performance improvement for the tiled version, due to cache effects. It may be relatively small, but is always there, so I thought it was not pure noise: lcs-a30c7480
> 
> On the CI system, it looks like the sequential non-tiled and tiled versions are not always ranked in the same order. But we could also adapt the text to explain that cache effects may or may not matter depending on the system (and I suspect that `BenchmarkTools` hides some of the effects by always working with a hot cache)

I see. Here is what it gives on my laptop:

lcs-5d4fafb7

So no appreciable difference between the sequential tiled vs. sequential non-tiled for me.

docs/src/examples/lcs/lcs.jl
```diff
@@ -0,0 +1,305 @@
...
+
+function init_lengths!(L, x, y)
+    @inbounds L[1,1] = 0
+    @inbounds for i in eachindex(x)
+        L[i+1, 1] = 0
+    end
+    @inbounds for j in eachindex(y)
+        L[1, j+1] = 0
+    end
+end
```
maltezfaria: I feel unsafe having `@inbounds` here since `L` is coming from outside. Is it
really important for performance? Also, I would probably have done:
```suggestion
function init_lengths!(L, x, y)
    L[1,:] .= 0
    L[:,1] .= 0
end
```

docs/src/examples/lcs/lcs.jl
```diff
@@ -0,0 +1,305 @@
...
+
+function fill_lengths!(L, x, y, ir=eachindex(x), jr=eachindex(y))
+    @inbounds for j in jr
+        for i in ir
+            L[i+1, j+1] = if x[i] == y[j]
+                L[i, j] + 1
+            else
+                max(L[i+1, j], L[i, j+1])
+            end
+        end
```
maltezfaria: Is the one liner obscure?
```suggestion
        for i in ir
            L[i+1, j+1] = x[i] == y[j] ? L[i, j] + 1 : max(L[i+1, j], L[i, j+1])
        end
```

maltezfaria: May I suggest also merging the loops?
```
for j in jr, i in ir
    ...
end
```

docs/src/examples/lcs/lcs.jl
```diff
@@ -0,0 +1,305 @@
...
+    end
+    String(subseq)
+end
+
+backtrack(L, x, y)
+
+# Wrapping everything into a function, we get
+
+function LCS(x, y)
+    L = Matrix{Int}(undef, 1+length(x), 1+length(y))
```
maltezfaria: Ok, I see it is "safe" to use `@inbounds` now. Sorry.

docs/src/examples/lcs/lcs.jl
maltezfaria: Turns out the algorithm is very simple, and well explained! Thanks for this. I will finish going over it later, and perhaps play around with the performance a bit myself.

docs/src/examples/lcs/lcs.jl
```diff
@@ -0,0 +1,305 @@
...
+
+plot(log_info, categories=["init", "tile", "backtrack"])
+
+# Here, we see for example that the run time is bounded by the length of the
+# critical path. Adding more threads would not help, and dividing the problem
+# into smaller chunks probably would not help much, either.
+#
+# Indeed, it appears that a non-negligible fraction of the time is spent in the
+# initialization of the array, which is mostly the memory allocation time for
+# the array `L`. This is a limitation of the current implementation, in which
```
maltezfaria: It seems from the code that the `init` does not take into account the allocation of
the matrix `L`.

docs/src/examples/lcs/lcs.jl
```diff
@@ -0,0 +1,305 @@
...
+                @R view(L, irange, jrange)
+                @W view(L, irange .+ 1, jrange .+ 1)
+                fill_lengths!(L, x, y, irange, jrange)
+            end label="tile ($kx, $ky)"
+        end
+    end
+
+    bt = DataFlowTasks.@spawn backtrack(@R(L), x, y) label="backtrack"
+    fetch(bt)
+end
```
maltezfaria:
```suggestion
function LCS_par(x, y, nx, ny)
    L = Matrix{Int}(undef, 1+length(x), 1+length(y))
    for (kx, irange) in enumerate(SplitAxis(eachindex(x), nx))
        DataFlowTasks.@spawn begin
            @W(view(L, irange, 1)) .= 0
        end label="init x ($kx,1)"
        for (ky, jrange) in enumerate(SplitAxis(eachindex(y), ny))
            if kx == 1
                DataFlowTasks.@spawn begin
                    @W(view(L, 1, jrange)) .= 0
                end label="init y ($ky)"
            end
            DataFlowTasks.@spawn begin
                @R view(L, irange, jrange)
                @W view(L, irange .+ 1, jrange .+ 1)
                fill_lengths!(L, x, y, irange, jrange)
            end label="tile ($kx, $ky)"
        end
    end

    bt = DataFlowTasks.@spawn backtrack(@R(L), x, y) label="backtrack"
    fetch(bt)
end
```

maltezfaria: I think this is correct, and it should be better than having a single
`init` block, and have everyone wait for that to start.

maltezfaria (commented):
I left a few comments, most of which are trivial. I did, however, suggest a rewrite of the `LCS_par` function which I may be better in that it exposes parallelism more quickly. 

If yo are OK with the suggested `LCS_par`, I think the text at the end needs to be changed a bit. 

Otherwise, LGTM. This is a nice example!

.github/workflows/CI.yml
ffevotte: Yes, I shared the same concern, which is why I put this back to the Documentation task. There, the relevant environment is already instantiated (and hopefully precompiled). And IIUC, it also runs on the self-hosted runner.

But it does increase the documentation build time, for an added 4% coverage. So is it worth it?

docs/run_examples.jl
ffevotte: Well, maybe `ispath` is not so aptly named: it checks whether the given path corresponds to an existing file.
In this case, this will be true for all examples following the naming convention `src/examples/$NAME/$NAME.jl`. This allows avoiding half backed examples which are already committed to the repository, but aren't yet in a runnable state.

docs/src/examples/blur-roberts/blur-roberts.jl
ffevotte: I guess it mostly has to do with my editor and syntax highlighting color scheme. I feel like it's easier on my eyes when the markdown parts of the file can be identified via comment markers, rather than only by a shaded color.

docs/src/examples/lcs/lcs.jl
ffevotte: Yes, I mostly followed the explanation given on the Wikipedia page, which is very well written indeed!

docs/src/examples/lcs/lcs.jl
ffevotte: Thanks, of course this is better!

docs/src/examples/lcs/lcs.jl
ffevotte: Yes, good ideas!

docs/src/examples/lcs/lcs.jl
ffevotte: Strange: I was positively sure I had tested this, and come to the conclusion that it didn't improve things. In my recollection, the first array write triggered the entire array allocation, which took more time than actually iterating over the first row + column.

But I re-tried it right now following your suggestion, and now it seems to work! Did I dream?

.github/workflows/CI.yml
maltezfaria: Let's keep it for now. If the build times eventually becomes annoyingly long, we can revisit this choice.

docs/run_examples.jl
maltezfaria: Got it, thanks.

docs/src/examples/lcs/lcs.jl
```diff
@@ -0,0 +1,297 @@
...
+x = collect("GAC");
+y = collect("AGCAT");
+
+# Let's build the $L$ array. For the sake of the example, we'll initially fill
+# it with $(-1)$ in order to indicate which values haven't been computed yet. The
+# `init_length!` function fills the first row and the first column in the
+# array. We also define a `display` helper function, which shows a pretty
+# representation of the $L$ array.
+
+function init_lengths!(L, x, y)
```
maltezfaria: Now the arguments `x` and `y` are unused...

ffevotte:
Thanks for the review! The last commit should account for all your comments.

The performance is indeed much better when parallelizing the initialization like you suggested. However, it's still not ideal when we create more tasks to expose more parallelism. For now, I (shamefully) bypassed the issue by slightly decreasing the number of tiles, which makes the critical path a limiting factor for the performance (arguably, it also helps reading the task graph)

maltezfaria (commented):
I want to make some minor edits, probably faster to do them directly. If you are OK with it, I will make them and push to this branch.

One not-so-minor thing that comes to my mind is that, much like the `cholesky` example, it may make sense to separate the allocation from the execution. That is, add a `LCS_par!` method. I think the speed-up will be much better in that case, and one could argue that you could do the `LCS` on several different pairs with same preallocated matrix. Are you OK with this?

ffevotte:
Please go ahead!

Likewise: can I directly push into #62 to make suggestions?

maltezfaria:
> Please go ahead!
> 
> Likewise: can I directly push into #62 to make suggestions?

Sure.

ffevotte:
> One not-so-minor thing that comes to my mind is that, much like the `cholesky` example, it may make sense to separate the allocation from the execution.

I considered this while writing the example, and eventually decided against. I think the big difference w.r.t the `cholesky` example is that here the only problem data are the two strings and the array is an internal working buffer. Whereas in `cholesky` the array *is* the data, so that the user necessarily has to allocate it somewhere before calling the factorization.

> That is, add a `LCS_par!` method. I think the speed-up will be much better in that case, and one could argue that you could do the `LCS` on several different pairs with same preallocated matrix. Are you OK with this?

Yes, if one has to perform LCS on several pairs of sequences of the same lengths, then one could indeed allocate a working buffer once, and re-use it. I have no idea whether this would be a real use-case.

But you're right, of course: the speed-up will probably be much better if we only measure it on the non-allocating parts of the algorithm. So maybe we could indeed add such an `LCS_par!` variant.

Do as you like, I'm fine with either way!

maltezfaria:
Ok, I think I am happy with this example. @ffevotte feel free to merge it if you think it is ready!

maltezfaria:
Sorry for so many typos in so few lines!

Regarding the implementation and one-offset, I think the algorithm was fine. In the loop you only `read` on `irange` and `jrange`, but you write on `irange.+1` and `jrange+1`, so you only need to make sure you have initialised the `irange/jrange` row/column. Am I wrong?

ffevotte:
> Regarding the implementation and one-offset, I think the algorithm was fine. In the loop you only `read` on `irange` and `jrange`, but you write on `irange.+1` and `jrange+1`, so you only need to make sure you have initialised the `irange/jrange` row/column. Am I wrong?

The algorithm looks like this:
```
function fill_lengths!(L, x, y, ir = eachindex(x), jr = eachindex(y))
    @inbounds for j in jr, i in ir
        L[i+1, j+1] = (x[i] == y[j]) ? L[i, j] + 1 : max(L[i+1, j], L[i, j+1])
    end
end
```
so for at any iteration `(i,j)`, we read from `L[i,j]`, `L[i+1,j]` and `L[i,j+1]`, and write to `L[i+1,j+1]`. So you really need to initialize the first column / row up to `irange .+ 1` / `jrange .+ 1`. It's actually relatively easy to see the bug in action: initializing `L` with very large values creates spurious very long common subsequences.

<br>

On another note, I really like the new ending: "try and see for yourself!" :smiley:

maltezfaria:
You are right, it was a bug and not a feature 😄 

I think I need some sleep... Merge this?

PR #61, opened 2023-10-24
Logging should wait for task graph to be empty

### LINKED ISSUE #59: Thread-safety of logging?
Because of the recent thread migration system in Julia, I don't think the following lines in the logging mechanics are thread-safe:

https://github.com/maltezfaria/DataFlowTasks.jl/blob/9e4c09319611f0b5663c00c59f6b7332da16cb27/src/dataflowtask.jl#L52-L56

The reason is that when a `Task` actually pushes into the `tasklogs` vector, it may live on a different thread than its `tid` variable. As such, two tasks on two different threads, but with the same value of `tid` could push concurrently to the `Vector{TaskLog}` used for logging, creating a race condition. 

The current design has `taskslogs` as vector of vectors of `TasksLog`s. After searching the code, most of the use of `tasklogs` actually just flattens that structure using `Iterators.flatten` in order to loop over all `TaskLog`s, so I am assuming that the original design was so that different tasks could just push to the vector-of-vectors structure without using any locks. But as mentioned above, this may no longer be *safe*. An alternative is to replace the vector-of-vectors simply by a `Channel{TaskLog}`. This may induce some overhead when the logging is active, but I doubt it will be anything major (also performance is not of uttermost importance when logging IMO).

### LINKED ISSUE #60: Logging should wait for task graph to be empty
Turns out some of the fails on the `CI`, such as:

- https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/6615534083/job/17967897823
- https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/6616191816/job/17970165311

are due to the fact that the logger does not check for underlying `TaskGraph` to be empty before starting. The following MWEs are representative of the possible issues with that.

## Example 1

A task is created before the `@log` block, but it is not yet finished when the logging starts. Since the task logs itself at the end of its execution, the `log_info` block below will contain two tasks instead of one:
```julia
import DataFlowTasks as DFT
A = Vector{Float64}(undef, 4)
t1 = DFT.@spawn (sleep(5); identity(@R A)) label = "Read"
log_info = DFT.@log let
    res = DFT.@spawn identity(@W A) label = "Write"
    fetch(res)
end
```

## Example 2

A somewhat contrived example arises when a task created before the logging block has finished executing, but has not yet been removed from the `DAG` (i.e. it is in the `FinishedChannel` waiting to be processed). If the logging block starts there, and a task spawned depends on the outside task just mentioned, then the logger will record a dependency on a task that it does not itself log! This yields the `KeyError` seen in the `CI` failures linked above. The code block below mimics this behaviour by manually deactivating and reactivating the task responsible for processing the `FinishedChannel`:

```julia
import DataFlowTasks as DFT
A = Vector{Float64}(undef, 4)
DFT.stop_dag_cleaner()
t1 = DFT.@spawn (identity(@R A)) label = "Read"
log_info = DFT.@log let
    res = DFT.@spawn identity(@W A) label = "Write"
    DFT.start_dag_cleaner()
    fetch(res)
end
```

maltezfaria:
- Fixes #60 by making sure the `TaskGraph` is empty before logging starts
- Adds a sanity check to address concerns with task migration during logging process (#59)

### DIFF
--- a/src/dataflowtask.jl
+++ b/src/dataflowtask.jl
@@ -34,11 +34,6 @@ mutable struct DataFlowTask
         tj = new(data, mode, TASKCOUNTER[], priority, label)
         addnode!(sch, tj, true)
 
-        # Store inneighbors if logging activated
-        _log_mode() &&
-            haslogger() &&
-            (inneighbors_ = [task.tag for task in inneighbors(sch.dag, tj)])
-
         deps = inneighbors(sch.dag, tj) |> copy
         tj.task = @task handle_errors() do
             for ti in deps
@@ -51,7 +46,7 @@ mutable struct DataFlowTask
             # Push new TaskLog if logging activated
             if _log_mode() && haslogger()
                 tid = Threads.threadid()
-                task_log = TaskLog(tj.tag, t₀, t₁, tid, inneighbors_, tj.label)
+                task_log = TaskLog(tj.tag, t₀, t₁, tid, [t.tag for t in deps], tj.label)
                 push!(_getloginfo().tasklogs[tid], task_log)
             end
             put!(sch.finished, tj)
--- a/src/dataflowtask.jl
+++ b/src/dataflowtask.jl
@@ -48,6 +48,9 @@ mutable struct DataFlowTask
                 tid = Threads.threadid()
                 task_log = TaskLog(tj.tag, t₀, t₁, tid, [t.tag for t in deps], tj.label)
                 push!(_getloginfo().tasklogs[tid], task_log)
+                if Threads.threadid() != tid
+                    @warn "Thread migration occurred while inserting a TaskLog"
+                end
             end
             put!(sch.finished, tj)
             return res
--- a/src/logger.jl
+++ b/src/logger.jl
@@ -121,12 +121,14 @@ function with_logging!(f, l::LogInfo)
     # taskgraph must be empty before starting, or we may log dependencies on
     # tasks that are not in the logger
     tg = get_active_taskgraph()
-    _wait = !isempty(tg)
-    msg = """logging requires an empty taskgraph to start. Waiting for pending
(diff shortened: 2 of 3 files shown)


src/logger.jl
```diff
@@ -117,6 +118,17 @@ end
...
 function with_logging!(f, l::LogInfo)
+    # taskgraph must be empty before starting, or we may log dependencies on
+    # tasks that are not in the logger
+    tg = get_active_taskgraph()
+    _wait = !isempty(tg)
+    msg = """logging requires an empty taskgraph to start. Waiting for pending
+    tasks to be completed...
+    """
+    _wait && (@warn msg; wait(tg))
+    _wait && (@warn "done waiting.")
```
ffevotte: Just a question of style, but is there a reason to avoid a (IMO) clearer construct like:
```julia
if !isempty(tg)
    msg = "waiting..."
    @warn msg
    wait(tg)
    @warn "done."
end
```

src/dataflowtask.jl
```diff
@@ -51,8 +46,11 @@ mutable struct DataFlowTask
             # Push new TaskLog if logging activated
             if _log_mode() && haslogger()
                 tid = Threads.threadid()
-                task_log = TaskLog(tj.tag, t₀, t₁, tid, inneighbors_, tj.label)
+                task_log = TaskLog(tj.tag, t₀, t₁, tid, [t.tag for t in deps], tj.label)
                 push!(_getloginfo().tasklogs[tid], task_log)
+                if Threads.threadid() != tid
+                    @warn "Thread migration occurred while inserting a TaskLog"
+                end
```
ffevotte, line 53: Faster than light edits!

ffevotte (approved):

PR #62, opened 2023-10-25
Expose `ExtendedLogInfo` and improve logging experience without add-ons

maltezfaria:
This `PR` moves some code from the Makie extension into the `logger.jl` file. This allows for *some* introspection of the `LogInfo` object without requiring any extra dependency. In particular

- `ExtendedLogInfo` objects can be constructed from a `LogInfo` object, and are now documented
- An example is added to the `Profiling` section of the docs to illustrate how the `ExtendedLogInfo` can be constructed and displayed
- Docstrings are added to `Gantt`

While I am still not fully happy with the logging interface, I think this `PR` provides a partial fix to #39, and IMO is enough for now.

PR #64, opened 2023-10-26
Simpler `describe(loginfo, categories)` API

ffevotte:
Following the changes in #62, I wanted to suggest an IMO simpler API to show detailed information about a `LogInfo` object. Instead of exposing `ExtendedLogInfo` and `extractloggerinfo` (which by side-effect also pulls in `Gantt`), I thought we could simply provide a function that extracts all the info under the hood and simply displays it.

I suggest using a function named `describe` for that (which is heavily inspired by [`StatsBase`](https://juliastats.org/StatsBase.jl/stable/scalarstats/#DataAPI.describe)), along the lines of:
```julia-repl
julia> log_info = DataFlowTasks.@log work(A, B)
julia> DataFlowTasks.describe(log_info; categories=["init", "mutate", "read"])
• Elapsed time           : 0.036
  ├─ Critical path       : 0.036
  ╰─ No-wait             : 0.006

• Run time               : 0.288
  ├─ Computing           :   0.051
  │  ├─ init             :     0.013
  │  ├─ mutate           :     0.034
  │  ├─ read             :     0.004
  │  ╰─ unlabeled        :     0.000
  ├─ Task insertion      :   0.000
  ╰─ Other (waiting)     :   0.237
```

Of course, all this comes from the assumption that `ExtendedLogInfo` instances aren't otherwise useful to the user. But maybe I'm wrong here?

If you like this idea, I can go ahead and update existing examples so that they also demonstrate the use of this new API.

[Documentation preview](https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR64/profiling/)

### DIFF
--- a/docs/src/profiling.md
+++ b/docs/src/profiling.md
@@ -55,7 +55,6 @@ profiling the compilation time, it is often advisable to perform a "dry run" of
 the code first, as done in the example below:
 
 ```@example profiling
-
 # Context
 A = ones(2000, 2000)
 B = ones(3000, 3000)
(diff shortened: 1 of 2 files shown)


maltezfaria (approved):
AFAIU `ExtendedLogInfo` is only used as an auxiliary structure in the code, so I think the changes proposed are fine. 

I like how we can now extract useful information from a run without needing to plot stuff now!

PR #67, opened 2023-10-29
New example: merge sort

### LINKED ISSUE #25: Tricky behavior with captured variables
When trying to implement new examples with DataFlowTasks, I encountered the following situation, which I find very tricky and hard to understand.

The example below is not so minimal, but it is not huge either, and it does something useful (solving a constant-speed transport equation in 1D) which is always nice IMO:

```julia
using Plots
using DataFlowTasks

function timestep!(u_new, u_old, α)
    for i in 2:lastindex(u_old)
        u_new[i] = u_old[i] - α * (u_old[i] - u_old[i-1])
    end
end

function solve!(uₙ, u₀, n, α)
    @assert iseven(n)

    # Two workspaces
    uₖ = uₙ          # solution at current timestep (known)
    uₖ₊₁ = zero(uₙ)  # solution at next timestep (to be computed with the FD scheme)

    copy!(uₖ, u₀)
    for k in 1:n
        # uncomment the variables in the let-block below to fix the problem
        let #uₖ₊₁=uₖ₊₁, uₖ=uₖ
            @dspawn timestep!(@W(uₖ₊₁), @R(uₖ), α) label="iter $k"
        end

        # !!! THIS IS THE TRICKY PART !!!
        # Swap both vectors:
        # - the newly computed solution becomes the known one
        # - the last known timestep will be re-used for the next time step
        (uₖ, uₖ₊₁) = (uₖ₊₁, uₖ)
    end

    # if the number of time steps is even, the last timestep will be in uₙ
    # (otherwise, we'd need one final copy from the temporary array uₖ₊₁ into uₙ)
    res = @dspawn @R(uₙ) label="result"
    fetch(res)
end
```

And here are some data to test it:
```julia
# Test case
N = 1000        # number of spatial meshes
c = 0.8         # advection speed
h = 1 / (N-1)   # spatial step
Δt = h          # time step (respects the CFL condition if c ≤ 1)
α = c*Δt/h      # convenient parameter for the scheme

# initial solution
x = 0:h:1
u₀ = (1 .+ sinpi.(5x)) .* (0.3 .< x .< 0.7)

# solution after 4 time steps
uₙ = similar(u₀)
solve!(uₙ, u₀, 4, α)
@assert uₙ != u₀
```

As it stands, it looks like the DAG is correct (at least it always was in my tests), but the computation almost always ends up with uₙ == u₀, which means that it *looks like* no computation was actually performed in the spawned tasks.

This is because `@dspawn timestep!(uₖ₊₁, uₖ, α)` makes a closure that captures the variables `uₖ₊₁` and `uₖ`, which are local to the function (not the `for` loop). This means that swapping the variables (i.e. re-binding them differently) at the end of each iteration can (and will) affect the behavior of the closures.

Putting a `let` block around the task creation works around the problem by making sure that the closures capture variables that are purely local to the `let` block.

I wonder whether we shouldn't make `@dspawn` & friends generate such a `let`-block automatically, because (1) this could become quite hard to debug, and (2) writing such `let`-blocks by hand is very tedious.

ffevotte:
Preview: https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR67/examples/sort/sort/


ffevotte:
This is (one of) the half-baked example I had lying around. It is interesting, in that it exhibits the kind of weird behavior described in #25.

This PR is now only a draft: it contains a first version of the code, as well as a few explanations. @maltezfaria if you like it, I can finish adding more explanations (including details about the long-awaited warning regarding captured variables).

ffevotte:
Now the code should be complete; the only missing part is a detailed explanation of the tricky behavior of captured variables.

ffevotte:
This should finally be ready for review!

maltezfaria:
@ffevotte I have gone through the new example, and made some small changes, mostly adding a bit more text here and there. Feel free to revert the commits if you disagree. 

One thing I decided to change was the logic behind the parallel merge function. The way things were done felt a bit hacky (e.g. having a fictitious index `k'`). IMO the simplest way to go about this is to have `parallel_merge!` which uses `Threads.@threads`:

```julia
function parallel_merge!(dest, left, right)
    ## Number of parts in which large blocks will be split
    P = min(8, ceil(Int, length(dest) / 65_536))
    ## Simple sequential merge for small cases
    if P <= 1
        merge!(dest, left, right)
        return dest
    end

    ## Split the arrays into `P` parts.
    idxs = split_indices(P, dest, left, right)
    ## Spawn one task per part
    Threads.@threads for p in 1:P
        let (iₚ, jₚ, kₚ) = idxs[p]
            @views merge!(dest[kₚ], left[iₚ], right[jₚ])
        end
    end
    return dest
end

```

The main `mergesort` function can then spawn a `DataFlowTask` when calling this `parallel_merge!` with the appropriate data dependencies:

```julia
...
DataFlowTasks.@spawn parallel_merge!(@W(dest), @R(left), @R(right)) 
...
```

 This works fine, and gives a similar speedup, but since we are creating `Task`s inside a `DataFlowTask`, the parallel trace is a bit messed up. In particular, as far as `DataFlowTasks` is concerned, the `parallel_merge!` is a single (kind of macro) task. I think this is a composability issue that we may want to address further down the road by at least adding some discussion to the docs. 

After some thought, and because of the logging limitations mentioned above, I decided to have what I think is a more natural solution, where the parallel merge function creates a task for `split_indices`, and then fetches its result before spawning the parallel tasks responsible of merging the sub-blocks. On my machine this turns out to be slightly slower, but feels *safer* in the sense that the data dependency is actually correct. What do you think?

maltezfaria:
@ffevotte I have gone through the new example, and made some small changes, mostly adding a bit more text here and there. Feel free to revert the commits if you disagree. 

One thing I decided to change was the logic behind the parallel merge function. The way things were done felt a bit hacky (e.g. having a fictitious index `k'`). IMO the simplest way to go about this is to have `parallel_merge!` which uses `Threads.@threads`:

```julia
function parallel_merge!(dest, left, right)
    ## Number of parts in which large blocks will be split
    P = min(8, ceil(Int, length(dest) / 65_536))
    ## Simple sequential merge for small cases
    if P <= 1
        merge!(dest, left, right)
        return dest
    end

    ## Split the arrays into `P` parts.
    idxs = split_indices(P, dest, left, right)
    ## Spawn one task per part
    Threads.@threads for p in 1:P
        let (iₚ, jₚ, kₚ) = idxs[p]
            @views merge!(dest[kₚ], left[iₚ], right[jₚ])
        end
    end
    return dest
end

```

The main `mergesort` function can then spawn a `DataFlowTask` when calling this `parallel_merge!` with the appropriate data dependencies:

```julia
...
DataFlowTasks.@spawn parallel_merge!(@W(dest), @R(left), @R(right)) 
...
```

 This works fine, and gives a similar speedup, but since we are creating `Task`s inside a `DataFlowTask`, the parallel trace is a bit messed up. In particular, as far as `DataFlowTasks` is concerned, the `parallel_merge!` is a single (kind of macro) task. I think this is a composability issue that we may want to address further down the road by at least adding some discussion to the docs. 

After some thought, and because of the logging limitations mentioned above, I decided to have what I think is a more natural solution, where the parallel merge function creates a task for `split_indices`, and then fetches its result before spawning the parallel tasks responsible of merging the sub-blocks. On my machine this turns out to be slightly slower, but feels *safer* in the sense that the data dependency is actually correct. What do you think?

ffevotte:
> After some thought, and because of the logging limitations mentioned above, I decided to have what I think is a more natural solution, where the parallel merge function creates a task for `split_indices`, and then fetches its result before spawning the parallel tasks responsible of merging the sub-blocks. On my machine this turns out to be slightly slower, but feels _safer_ in the sense that the data dependency is actually correct. What do you think?

Yes, I tried this and originally rejected it because I wanted to get the best performance. But you're right, let's go for the safe way now (especially for a tutorial!) and maybe update the example later if/when we get a new API allowing to spawn tasks within tasks...

maltezfaria:
The performance difference is quite minimal actually, specially after adding the type declaration to the fetched `idxs` variable, so I am glad we decided to go with this solution 👍 

Thanks for the nice example and illustrations by the way!

PR #71, opened 2023-11-07
Update Cholesky example

### LINKED ISSUE #70: Update `Cholesky` example with `LoopVectorization` kernels
Now that the `CI` runs the examples on a workstation, we could update the `cholesky` example to include the full-fledge implementation using `LoopVectorization`. We can then compare that to `MKL` and see how it fares.

maltezfaria:
Doc preview:

https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR71/examples/cholesky/cholesky/


maltezfaria:
@ffevotte I think this is almost ready for a review, I will let you know. Some comments before I forget:

- Turns out your implementation was already quite good, and there is not need to use `LoopVectorization` or anything else to get decent performance. 
- I changed the `CI` so that
    - It stops running the tests on `TiledFactorization`. I will deprecate that package now. 
    - Only run the examples with code coverage if we are on main. This seems like a reasonable compromise. 
- Using `MKL` turned out to be a bit tricky (I failed on my initial attempt). Also, it introduces some heavy dependency. I think it is fair to make the comparison only to `BLAS` version shipped by default with Julia. I did write a comment that if available one should definitely try `MKL` 😄 
-  Running the examples on `8` cores is heavily underusing the machine we have (which has 20 cores). Should we change it to `16`? One issue could be the visual aspect of the plots, which may be overwhelming. 

Left to do:
- [x] Update the README
- [x] Tweak the plots a bit. I think we are running too many benchmarks.

ffevotte:
> * Turns out your implementation was already quite good, and there is not need to use `LoopVectorization` or anything else to get decent performance.

Cool! But AFAIR it's not my implementation: it's yours!



> I changed the `CI` so that
> 
>     * It stops running the tests on `TiledFactorization`. I will deprecate that package now.

That's good too IMO.
 
>     * Only run the examples with code coverage if we are on main. This seems like a reasonable compromise.

It'll probably make it a bit harder to track coverage issues in PRs, but saving resources is a noble endeavor!



> * Using `MKL` turned out to be a bit tricky (I failed on my initial attempt). Also, it introduces some heavy dependency. I think it is fair to make the comparison only to `BLAS` version shipped by default with Julia. I did write a comment that if available one should definitely try `MKL` 😄

100% agree



> * Running the examples on `8` cores is heavily underusing the machine we have (which has 20 cores). Should we change it to `16`? One issue could be the visual aspect of the plots, which may be overwhelming.

AFAIK, the way we do things, all examples run with the same number of cores. And I worry that running `lcs` or `sort` on more than 8 cores show relarively bad performance (esp. with the current problem sizes). So I'd keep things like they currently are.

maltezfaria:
@ffevotte This is ready for a review. 

p.s. I changed the number of the threads to `16`, feel free to revert that if you prefer how the figures look with `8`.

ffevotte:
That's very nice! Impressive that this naive implementation already outperforms OpenBLAS (by a fair margin, with 16 threads!)

I deleted some comments about the performance of blur-roberts, which were not really relevant anymore. I also tried increasing the maximum number of "parts" for parallel merges in the sort example, but it does not always work (the performance varies a lot from run to run).

Feel free to merge if you agree with these minor changes.

And thanks!

ffevotte:
Also: what is `dependabot` useful for?

maltezfaria:
> Also: what is `dependabot` useful for?

In this case it is supposed to update the GitHub actions for you:

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

maltezfaria:
> That's very nice! Impressive that this naive implementation already outperforms OpenBLAS (by a fair margin, with 16 threads!)
> 
> I deleted some comments about the performance of blur-roberts, which were not really relevant anymore. I also tried increasing the maximum number of "parts" for parallel merges in the sort example, but it does not always work (the performance varies a lot from run to run).
> 
> Feel free to merge if you agree with these minor changes.
> 
> And thanks!

It is also faster on 8 threads... I just felt that running a parallel program using 8 threads on a machine with 20 cores begs for some questions 🤔 

I looked over the new figures now, and although now the speed up is suboptimal, it seems we are always close to either the  two lower bounds (critical time and no-wait time). So maybe this is fine?

I am gonna make some small changes:

- Add a *GFlops/seconds* plot next to the runtime. 
- Add a section to the examples on the hardware specifications, instead o having this as a subsection in each example

maltezfaria:
Ok, I think I am (finally) ready to merge this. At the end, I decided for `8` threads. Most of the problems in the examples are too small for 16 threads, and increasing the problem size increases the build time for these examples, which is already quite large!

maltezfaria:
BTW, I am gonna go with a *squash and merge* here (too many useless commits)

PR #77, opened 2023-11-08
Expand discussion on nesting tasks and limitations

### LINKED ISSUE #40: Document troubleshooting and debugging strategies

maltezfaria:
Still WIP, but the idea is to explain a bit better the current limitations. In particular, why *nesting* is somewhat tricky, and what are the current workarounds. Maybe this will also help us decide on a proper way forward, such as the nesting of `TaskGraph`s and merging of `LogInfo` objects.


maltezfaria:
@ffevotte I think this is ready for a first review. I found it quite tricky to explain the problem with nested tasks, but I tried 😅 

https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR77/troubleshooting/#nested-tasks

maltezfaria:
While I was at it, I also added a small *suggested workflow* section with the experience described in [#40 ](https://github.com/maltezfaria/DataFlowTasks.jl/issues/40#issuecomment-1274549448)

PR #80, opened 2023-11-17
Allow for `Task`s as a data-dependency in a `DataFlowTask`

### LINKED ISSUE #79: Fetching inside a `DataFlowTask`
Suppose we need to fetch a `DataFlowTask` inside another:

```julia
function foo()
    d1 = @dspawn (sleep(1); rand(10)) label="sleep"
    d2 = @dspawn (fill!(fetch(d1),0)) label="fill"
    sum(fetch(d2))
end
```

This will work, and produce the correct result, but the logging is completely wrong since 
- from the `taskgraph` point of view, `d2` is independent of `d1`
- `d2` is logged as taking about 1 second. 

Ideally, we would have a `@RW(d1)` inside of `d2`, and that would insert an edge between `d2` and `d1`. How to do that?

maltezfaria:
An attempt to resolve #79.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ JuliaSysimage.dll
 Manifest.toml
 TiledFactorization
 docs/.DS_Store
+docs/slides/**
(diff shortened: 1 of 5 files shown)


maltezfaria:
LGTM

PR #81, opened 2023-11-20
Implicitly add a "data dependency" between each `DataFlowTask` and its underlying `Task`

ffevotte:
This is another way of handling #79, in which implicit "data dependencies" are registered between each `DataFlowTask` and its underlying `Task`. It's then possible to handle such dependencies with the regular, `memory_overlap`-based mechanism.

I feel like this is essentially equivalent to the current implementation, but perhaps a bit more straightforward in terms of implementation. @maltezfaria what do you think?

Preview doc: https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR81/

### DIFF
--- a/src/dataflowtask.jl
+++ b/src/dataflowtask.jl
@@ -55,6 +55,8 @@ mutable struct DataFlowTask
             put!(sch.finished, tj)
             return res
         end
+        tj.data = (tj.data..., tj.task)
+        tj.access_mode = (tj.access_mode..., READWRITE)
         return tj
     end
 end
@@ -115,9 +117,7 @@ end
 
 @noinline function _data_dependency(datai, modei, dataj, modej)
     for (di, mi) in zip(datai, modei)
-        (di isa Task) && continue # Tasks are handled differently
         for (dj, mj) in zip(dataj, modej)
-            (dj isa Task) && continue # Tasks are handled differently
             mi == READ && mj == READ && continue
             if memory_overlap(di, dj)
                 return true
@@ -140,6 +140,7 @@ arguments to allow for appropriate inference of data dependencies.
 """
 function memory_overlap(di, dj)
     (isbits(di) || isbits(dj)) && return false
+    (di isa Task || dj isa Task) && return di === dj
     @warn """using fallback `memory_overlap(::Any,::Any) = true`. Consider implementing
     `DataFlowTasks.memory_overlap(::$(typeof(di)),::$(typeof(dj)))`.
     """
--- a/src/dag.jl
+++ b/src/dag.jl
@@ -179,7 +179,6 @@ and outgoing edges are updated.
 function update_edges!(dag::DAG, nodej)
     transitively_connected = dag._buffer
     empty!(transitively_connected)
-    hastask = any(x -> isa(x, Task), data(nodej))
     # update dependencies from newer to older and reinfornce transitivity by
     # skipping predecessors of nodes which are already connected
     for (nodei, _) in Iterators.reverse(dag)
@@ -192,13 +191,7 @@ function update_edges!(dag::DAG, nodej)
         (ti ∈ transitively_connected) && continue
         # tasks are handled differently when they appear in the data in that
         # they are checked directly agains the nodej.task field
-        dep = false
-        if hastask
-            for d in data(nodej)
-                d === nodei.task && (dep = true; break)
-            end
-        end
-        dep || data_dependency(nodei, nodej) || continue
+        data_dependency(nodei, nodej) || continue
         addedge!(dag, nodei, nodej)
         update_transitively_connected!(transitively_connected, nodei, dag)
     end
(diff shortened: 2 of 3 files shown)


src/dataflowtask.jl
```diff
@@ -55,6 +55,8 @@ mutable struct DataFlowTask
             put!(sch.finished, tj)
             return res
         end
+        tj.data = (tj.data..., tj.task)
+        tj.access_mode = (tj.access_mode..., READWRITE)
```
ffevotte, line 59: This is a bit uglier than I'd like, and raises the question: is there an interest in having
```
    data::Tuple
    access_mode::NTuple{<:Any,AccessMode}
```
instead of something like
```
    data::Vector{Any}
    access_mode::Vector{AccessMode}
```

Anecdotally, this would make the implicitly added `DataFlowTask`->`Task` dependency a simple `push!`, but perhaps more importantly this would make sure `DataFlowTask` only has concretely typed fields. Wouldn't that be more desirable than storing `Tuple`s of unspecified lengths?

maltezfaria:
Your idea seems indeed simpler and better! Let's go with it.

src/dataflowtask.jl
maltezfaria: As for having the `Tuple` in the fields of `DataFlowTask`s, this was a deliberate (but possibly bad) choice. The rationale was that we could have only one dynamic dispatch when calling `data_dependency`  since we first unpack the tuple, and then call an internal (type stable?) function with an explicit `@noinline`:

https://github.com/maltezfaria/DataFlowTasks.jl/blob/7744ef625ea342b0e602acf7a67e143c5547c5d6/src/dataflowtask.jl#L116

In most applications of DFT that I tried there are not that many concrete types of `data` that you create, so this idea of *unpacking and dispatching* seemed reasonable. The option of having a vector means that for each pair `(datai,dataj)` you do a dynamic dispatch right?

In any case, your idea of having a `Vector` is simpler, but perhaps putting some benchmarks in place (using e.g. [PkgBenchmarks](https://github.com/JuliaCI/PkgBenchmark.jl) ) is needed before committing to such a change. Open an issue?

ffevotte: > In most applications of DFT that I tried there are not that many concrete types of data that you create, so this idea of unpacking and dispatching seemed reasonable. The option of having a vector means that for each pair (datai,dataj) you do a dynamic dispatch right?

Yes, so with a vector-based approach (V), `_data_dependency(datai, dataj)` dispatches statically, and does not need to be specialized. But inside it, every call to `memory_overlap(di, dj)` dispatches dynamically.

In the current version with Tuples (T), it's the other way around: several specializations of `_data_dependency` have to be compiled, and then called via a dynamic dispatch. Inside each one, however, all calls to `memory_overlap(di, dj)` hopefully dispatch statically.

I'd say approach (V) is more compile-time-friendly, but approach (T) is more run-time efficient. I doubt it would matter in practice though, but you're right: we'd need benchmarks to check this (and more generally track the performance impact of our commits)

maltezfaria: Yes, my impression was that (T) would be preferable for speed (i.e. run-time efficiency). That seems indeed to be the case, at least for the following simple example:

```julia
# Current tuple-style
A = rand(2); B = rand(2); C = rand(2)
di = DataFlowTasks.@dtask (@RW(A);@RW(B); @RW(C))
dj = DataFlowTasks.@dtask (@RW(C);@RW(B); @RW(A))
@btime DataFlowTasks.data_dependency($di,$dj) # 30.118 ns (0 allocations: 0 bytes)

# Vector-style
datai = Vector{Any}([d for d in di.data])
dataj = Vector{Any}([d for d in dj.data])
modei = [m for m in di.access_mode]
modej = [m for m in dj.access_mode]
@btime DataFlowTasks._data_dependency($datai,$modei,$dataj,$modej) # 77.168 ns (0 allocations: 0 bytes)
```

Of course, it may very well be that this speed difference is utterly irrelevant, but for now I would feel safer not changing it. I am gonna open an issue for this.

maltezfaria: Of course, things could get out of hand with the current approach if too many `data` types are created, as that would require compiling lots of `_data_dependency` methods. But I doubt that is a real concern...

maltezfaria:
If we can agree on not changing the handling of the `data` field for now, I think we can merge this PR.

src/dataflowtask.jl
maltezfaria: For posteriority, here is an even more dramatic case with more dynamic dispatches:

```
A =  rand(2); B = rand(2); C = rand(2)
A2 = rand(2); B2 = rand(2); C2 = rand(2)
di = DataFlowTasks.@dtask (@RW(A);@RW(B); @RW(C))
dj = DataFlowTasks.@dtask (@RW(A2);@RW(B2); @RW(C2))
@btime DataFlowTasks.data_dependency($di,$dj) # 33.693 ns (0 allocations: 0 bytes)

datai = Vector{Any}([d for d in di.data])
dataj = Vector{Any}([d for d in dj.data])
modei = [m for m in di.access_mode]
modej = [m for m in dj.access_mode]

@btime DataFlowTasks._data_dependency($datai,$modei,$dataj,$modej) # 224.060 ns (0 allocations: 0 bytes)
```

ffevotte:
> If we can agree on not changing the handling of the `data` field

Agreed!

PR #82, opened 2023-11-20
Cafe julia presentation

ffevotte:
We probably won't merge this PR (instead creating an orphan branch or something like this), but I'm opening it nonetheless so that we can use this space to discuss the contents of the presentation.


maltezfaria:
My schedule is a bit crazy this week, but I will try to work on it tonight to:

- [x] Add the simple example as a ~~`Pluto`~~`Jupyter` notebook
- [ ] Simplify the discussion of the logging part (remove the slides with examples, which are no longer necessary once we introduce them earlier)
- [ ] Some minor fixes

I will push the changes to this branch in any case.

ffevotte:
I already started working on the small example.

After trying one more time Pluto, I didn't manage to use it correctly for this (especially I couldn't control the number of threads and had difficulties managing the environment to use a development version of DFT). I did however manage to make things work with a Jupyter notebook (even the slideshow features, which caused me some trouble last time I tried due to an incompatibility with GraphViz).

Anyways, I'd be very fine with a switch to Pluto if you know how to make it work, but I'm going to commit my Jupyter notebook to at least save you the trouble of extracting the relevant pieces of code from the more complete example.

You'll see that for this small example, I also switched to a recursive, top-down approach. IMO, this has the advantage of being simpler to understand (and closer to the textbook algorithm), while also preventing tricky capture behavior (since each stack frame introduces its own bindings).

ffevotte:
PS: for a preview of the Jupyter-based slideshow, you can download [`sort.slides.html`](https://github.com/maltezfaria/DataFlowTasks.jl/raw/cafe-julia-presentation/docs/slides/cafe-julia-2023/sort.slides.html) and open it with a web browser

maltezfaria:
> I already started working on the small example.
> 
> After trying one more time Pluto, I didn't manage to use it correctly for this (especially I couldn't control the number of threads and had difficulties managing the environment to use a development version of DFT). I did however manage to make things work with a Jupyter notebook (even the slideshow features, which caused me some trouble last time I tried due to an incompatibility with GraphViz).
> 
> Anyways, I'd be very fine with a switch to Pluto if you know how to make it work, but I'm going to commit my Jupyter notebook to at least save you the trouble of extracting the relevant pieces of code from the more complete example.
> 
> You'll see that for this small example, I also switched to a recursive, top-down approach. IMO, this has the advantage of being simpler to understand (and closer to the textbook algorithm), while also preventing tricky capture behavior (since each stack frame introduces its own bindings).

Thanks! I will start from your Jupyter notebook and give Pluto a try.

maltezfaria:
I just added the `Pluto` version of the `Jupyter` notebook. I haven't played with their slide system yet, but I am wondering if it is worth it. The slides you had from `Jupyter` look good enough to me, so maybe lets just go with that.

ffevotte:
With the last commit:

- the Jupyter notebook contains everything needed to setup a temporary environment, so that people can download and run it without having to think twice about what to install and where
- the Pluto notebook has a "present" button as well as intermediate titles (which are needed to delimit slides)

All in all, both versions should now provide the same features, so it's now simply a matter of deciding what we prefer.

maltezfaria:
> With the last commit:
> 
> * the Jupyter notebook contains everything needed to setup a temporary environment, so that people can download and run it without having to think twice about what to install and where
> * the Pluto notebook has a "present" button as well as intermediate titles (which are needed to delimit slides)
> 
> All in all, both versions should now provide the same features, so it's now simply a matter of deciding what we prefer.

IMO `Jupyter` looks nicer in presentation mode, so I suggest we go with that (it is also the same format we use for the other notebooks). 

I will work a bit on the `documentation` branch to 
- add the setup header for the example notebooks we have, similar to what you did here. 
- proof-read the new sections
- add the discussion on limitations of logging for tasks that `yield`

Anything else you would like to add? We should probably tag a new version after that 🥳

ffevotte:
> IMO `Jupyter` looks nicer in presentation mode, so I suggest we go with that (it is also the same format we use for the other notebooks).

OK



> * add the setup header for the example notebooks we have, similar to what you did here.

You probably have already thought of that, but just in case: I think these headers should only be added to the notebooks, so that (1) documentation previews reflects the state of the code in the PR (rather than `main`), and (2) code coverage still collects data from documentation builds.

I'm not sure how to do that, though. Up to now, we've always considered our Literate scripts to be directly runnable (e.g. in `docs/run_examples`), but we might now need to use `Literate.script` to generate runnable scripts.

maltezfaria:
> > IMO `Jupyter` looks nicer in presentation mode, so I suggest we go with that (it is also the same format we use for the other notebooks).
> 
> OK
> 
> > * add the setup header for the example notebooks we have, similar to what you did here.
> 
> You probably have already thought of that, but just in case: I think these headers should only be added to the notebooks, so that (1) documentation previews reflects the state of the code in the PR (rather than `main`), and (2) code coverage still collects data from documentation builds.
> 
> I'm not sure how to do that, though. Up to now, we've always considered our Literate scripts to be directly runnable (e.g. in `docs/run_examples`), but we might now need to use `Literate.script` to generate runnable scripts.

I am working on it now. I think it suffices to create a preamble with lines that start with `#nb`. That way:
1. We can still execute the script as a plain Julia script
2. The preamble only goes into the notebook generation

maltezfaria:
I think I am done adding the major points I wanted to the slides. Should we discuss how to split the sections?

maltezfaria:
Also, should we tag `0.2` prior to the talk?

ffevotte:
> I think I am done adding the major points I wanted to the slides. Should we discuss how to split the sections?

Yes, I think we should. But I'm not available today... Tomorrow?

ffevotte:
I'm re-using the same branch for the JuliaCon local slides, under the directory `docs/slides/juliacon-local-2023`.

maltezfaria:
> I'm re-using the same branch for the JuliaCon local slides, under the directory `docs/slides/juliacon-local-2023`.

Sounds good, I will keep an eye on it. Here is the rough mental outline I had after we discussed today:

- An intro similar to what we had, minus the parallel merge example. Probably taking 10 to 15 min. 
- Entirely skip the implementation detail section. Move that to backup slides. 
- Focus on the cholesky example for the remainder of the talk (say to 10 to 15 minutes). IMO it should include:
  - why this example is a good benchmark (e.g. a good arithmetic intensity since you do $n^3$ flops for $n^2$ bytes, so you can expect to gain from "good" parallelism since you are compute-bound). For some of the other examples we have in the docs we are most likely memory-bound. 
  - A discussion of the tiled algorithm. It is not hard, but we need to show where it comes from. I would map each operation to the corresponding julia function at some point (e.g. `mul!`, `cholesky!`, `ldiv!`). Discuss that we call *OpenBLAS* for simplicity, but having those kernels efficiently done in `julia` is relatively simple if one uses `LoopVectorization`. 
  - Then show how to parallelize it using DFT
  - Explore the `@log` information, and how to read/interpret it
  - Comparison to OpenBLAS

I will try to find time in the next few days to work on this as well. Thanks!

maltezfaria:
@ffevotte I am going over the presentation now. The modified version of the example does:

```julia
@dspawn A \ @R(b)
```

I think it is better to have `@R(A) \ @R(b)` since then you can later have other tasks that access `A` (either before or after). One issue is that this change may require tweaking with the `tikz` graph that explains the basic idea. Worth it?

ffevotte:
> One issue is that this change may require tweaking with the `tikz` graph that explains the basic idea. Worth it?

Yes, I considered this, but I didn't manage to convince myself that it would be worth it. Especially for an example that is meant to be as simple as possible.

I actually tried to find a non-trivial operation working on the whole array, which could not be easily decomposed for each subarray and which would not involve more data... But I didn't have any idea.

ffevotte:
@maltezfaria I really like your animations for the Cholesky example. I'm only worried that the yellow font might be hard to read in the conference room. May I go ahead and change it to something else (maybe the green we have in other parts of the slides)?

maltezfaria:
> @maltezfaria I really like your animations for the Cholesky example. I'm only worried that the yellow font might be hard to read in the conference room. May I go ahead and change it to something else (maybe the green we have in other parts of the slides)?

I just pushed some changes. Please, do feel free to change the colors (I am really bad at choosing colors 😞 ).

I thought it was easier to explain it using the lower triangular part (fewer transposes involved), so I went ahead and did that.

maltezfaria:
> > One issue is that this change may require tweaking with the `tikz` graph that explains the basic idea. Worth it?
> 
> Yes, I considered this, but I didn't manage to convince myself that it would be worth it. Especially for an example that is meant to be as simple as possible.
> 
> I actually tried to find a non-trivial operation working on the whole array, which could not be easily decomposed for each subarray and which would not involve more data... But I didn't have any idea.

I see... not a big deal to leave it without. Oh, I just remembered that I started changing `A \ @R(b)` to `@R(A) \ @R(b)` locally on my version, and I just committed that, sorry! Should I go ahead and fix it or do you prefer to do it yourself?

ffevotte:
OK, I did both tweaks:
- `cholesky!` is now associated to a green color
- I removed `@R(A)` from the code in the animation (I left it in the first slide though, because it's more consistent to have it; although it's a bit inconsistent, I think this follows some form of "least surprise principle")

Now I don't plan on changing anything more

maltezfaria:
I am fixing some issues with the new `cholesky` pseudocode now. Give me a few minutes.

maltezfaria:
This LGTM, feel free to send it!

ffevotte:
Done! 23:53, we're even a bit early!

ffevotte:
Now that we're done with this, I thought we could:
- put the sources for the slides in a dedicated orphan branch (`talks`?), along with the resulting PDF slideshows
- maybe add a link to the slideshows in a dedicated page in the documentation

@maltezfaria Any toughts about this?

maltezfaria:
> Now that we're done with this, I thought we could:
> 
> * put the sources for the slides in a dedicated orphan branch (`talks`?), along with the resulting PDF slideshows
> * maybe add a link to the slideshows in a dedicated page in the documentation
> 
> @maltezfaria Any toughts about this?

I think these are both good ideas 😃

ffevotte:
I added the sources for the slides in a new [`talks`](https://github.com/maltezfaria/DataFlowTasks.jl/tree/talks) orphan branch. A basic `README.md` file links the available presentations as pdf.

The ["Going Further" section](https://github.com/maltezfaria/DataFlowTasks.jl#going-further-examples-and-performance)  in the main `README` points to that branch as an additional source of documentation.

@maltezfaria Unless you want to further discuss this, please feel free to close this PR

maltezfaria:
LGTM. Closing and deleting this branch. Thanks!

PR #84, opened 2023-11-21
Documentation updates

maltezfaria:
Doc preview:

https://maltezfaria.github.io/DataFlowTasks.jl/previews/PR84/


maltezfaria:
Transforming the notebooks for the examples that we have in the docs into relocatable (standalone) files is turning out to be more subtle than I thought. 

I can add a sort of preamble in each of the `someexample.jl` file like so:

```julia
#nb Pkg.activate(temp=true)
#np Pkg.add(name="DataFlowTasks", rev="main")
## ... add other dependencies needed for the example
```

IMO this is not very satisfactory, as we will have the documentation version begin executed with the current commit, while the notebook will use the version in `main`.  I could try to add the exact commit, but this comes with its own problems. 

@ffevotte Do we really intend for people to just *download* the `ipynb` version of the examples and run them from wherever? If so, how do you think we should we go about it? Every solution I tried seems hacky 😕

ffevotte:
> @ffevotte Do we really intend for people to just _download_ the `ipynb` version of the examples and run them from wherever? If so, how do you think we should we go about it? Every solution I tried seems hacky 😕

I'll try and think of something

>  I could try to add the exact commit, but this comes with its own problems.

Are you thinking about something specifically?

maltezfaria:
> Are you thinking about something specifically?

Besides the fact that the implementation of something like this (i.e. adding the specific commit in the notebook file) will require adding a pre-processing step to `Literate` (so some more coding), the main issue I see is that this will make sense on the *CI*, so we probably would have to change the `make.jl` for the documentation to only run the notebooks on the `CI`.

maltezfaria:
> > Are you thinking about something specifically?
> 
> Besides the fact that the implementation of something like this (i.e. adding the specific commit in the notebook file) will require adding a pre-processing step to `Literate` (so some more coding), the main issue I see is that this will make sense on the _CI_, so we probably would have to change the `make.jl` for the documentation to only run the notebooks on the `CI`.

Turns out it is not as hard as I thought to add to the notebook the commit of ~~the branch~~ revision that generated it. So I guess the only objection that remains is what does this mean to run such notebooks locally and/or in a dirty git state? Maybe not running the notebooks locally is OK after all, since they are not intended to be run on our controlled environment anyways 🤔

maltezfaria:
Ok, it seems like I just needed to say the "issue" out loud. I think the solution of adding the specific commit and disabling the execution of the notebooks if not on the `CI` seems reasonable. ~~I will implement that for the `cholesky.jl` case and we decide if it is worth doing it for the others.~~

ffevotte:
Yes, so I'm currently trying something that's almost exactly what you did, except it uses a pre-processing to add the whole setup cell. So that we can enable the pre-processing on CI to insert a setup with a specific git commit, but not activate any pre-processing set in local documentation builds (so that we can still look at the generated notebooks if we want)

maltezfaria:
> Yes, so I'm currently trying something that's almost exactly what you did, except it uses a pre-processing to add the whole setup cell. So that we can enable the pre-processing on CI to insert a setup with a specific git commit, but not activate any pre-processing set in local documentation builds (so that we can still look at the generated notebooks if we want)

This sounds even nicer... Feel free to push here if you have something working.

ffevotte:
I think this is finally working as expected. Cf the [cholesky notebook](https://nbviewer.org/github/maltezfaria/DataFlowTasks.jl/blob/gh-pages/previews/PR84/examples/cholesky/cholesky.ipynb)

Documenter seems to be a little confused by all these environment switches. Maybe it's not a good idea to generate all notebooks in the same process as the documentation. But things seem to be working now, and we can always change this later if needed.

maltezfaria:
Thanks! I will make some minor changes here and push.

p.s. What do you mean by *documenter seems to be a little confused*?

maltezfaria:
@ffevotte Should we merge? This seems to work, and improving the doc is a never-ending process, but at some point we just need to merge it.

maltezfaria:
🥳

PR #85, opened 2023-11-23
Use upstream version of reverse iterator for `OrderedDict`

maltezfaria:
Support for reverse iteration of a `OrderedDict` was added in https://github.com/JuliaCollections/OrderedCollections.jl/pull/108. This `PR` removes the now-deprecated code in `utils.jl`, and updates the `compat` entry of `OrderedCollections`.


maltezfaria:
**Note**:  I accidentally pushed directly into `main` initially instead of this branch, and to avoid force-pushing into `main` I did a revert. Git history is a bit ugly because of that 😞

PR #94, opened 2024-06-07
Update README.md

camilodlt:
Pkd => Pkg typo


camilodlt (commented):
Since its now registered

PR #96, opened 2025-02-12
__precompile__() on the main module and the extensions

camilodlt:
Hello :), 

I use DataFlowTasks.jl in one of my packages and I need to precompile it, however, I can't precompile it if it uses a package that has __precompile__(false) like the current version of DataFlowTasks.jl 

I think I put all the code that is needed for making the precompilation work. I changed the `PROJECT_ROOT` global to a constant `Ref{String}` that is updated when the package is __init__. 
I made a "hidden" method `_get_dataflowtasks_root` which just returns the value of the ref. I use that method everywhere where `PROJECT_ROOT` was used. 

I ran all the tests on my computer and got all green. 
I tested them on my machine which has the following `versioninfo()` and was able to precompile it: 
```
julia> versioninfo()
Julia Version 1.10.5
Commit 6f3fdf7b362 (2024-08-27 14:19 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, skylake)
Threads: 1 default, 0 interactive, 1 GC (on 12 virtual cores)
```

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -28,7 +28,7 @@ end
 for example in ["cholesky", "blur-roberts", "lcs", "sort"]
     println("\n*** Generating $example example")
     @time begin
-        dir = joinpath(DataFlowTasks.PROJECT_ROOT, "docs", "src", "examples", example)
+        dir = joinpath(DataFlowTasks._get_dataflowtasks_root(), "docs", "src", "examples", example)
         src = joinpath(dir, "$(example).jl")
         Literate.markdown(src, dir)
         draft || Literate.notebook(src, dir; preprocess = insert_setup)
@@ -37,7 +37,7 @@ end
 
 # generate readme
 println("\n*** Generating README")
-@time cd(joinpath(DataFlowTasks.PROJECT_ROOT, "docs", "src", "readme")) do
+@time cd(joinpath(DataFlowTasks._get_dataflowtasks_root(), "docs", "src", "readme")) do
     src = joinpath(pwd(), "README.jl")
 
     # Run code
--- a/ext/DataFlowTasks_GraphViz_Ext.jl
+++ b/ext/DataFlowTasks_GraphViz_Ext.jl
@@ -1,3 +1,4 @@
+__precompile__()
 module DataFlowTasks_GraphViz_Ext
 
 using GraphViz
--- a/ext/DataFlowTasks_GraphViz_Ext.jl
+++ b/ext/DataFlowTasks_GraphViz_Ext.jl
@@ -1,4 +1,3 @@
-__precompile__()
 module DataFlowTasks_GraphViz_Ext
 
 using GraphViz
--- a/ext/DataFlowTasks_Makie_Ext.jl
+++ b/ext/DataFlowTasks_Makie_Ext.jl
@@ -1,3 +1,4 @@
+__precompile__()
 module DataFlowTasks_Makie_Ext
 
 using Makie
--- a/ext/DataFlowTasks_Makie_Ext.jl
+++ b/ext/DataFlowTasks_Makie_Ext.jl
@@ -1,4 +1,3 @@
-__precompile__()
 module DataFlowTasks_Makie_Ext
 
 using Makie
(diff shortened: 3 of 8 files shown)


maltezfaria:
Thanks for the PR! It would be great to make DataFlowTasks pre-compilation friendly. 

Any idea why the CI is [failing on 1.11](https://github.com/maltezfaria/DataFlowTasks.jl/actions/runs/13296804253/job/37130510784?pr=96)? The error is

```julia
ERROR: LoadError: InitError: MethodError: Cannot `convert` an object of type Nothing to an object of type String
The function `convert` exists, but no method is defined for this combination of argument types.
```

camilodlt:
Hello again, so it failed on 1.6.7 because [global constants can not be typed on that version ](https://discourse.julialang.org/t/julia-1-7-and-below-do-not-have-constant-type-globals-julia-1-8-implemented-typed-globals/1989) and it failed on 1.11 because [we can not use `pkdir` inside the init in that version](https://github.com/JuliaLang/julia/issues/56077).

To solve it, I made a static check on the version and define a typed or untyped global constant accordingly. I left PROJECT_ROOT as a ref however. 
For the second issue, the solution was to get the `pkgdir` at precompilation time which I don't think will pose a problem and, in a way, is what you were doing before.

I removed the `__precompile__` statements from the extension as I'm not sure they are needed. 

In my computer it passes tests on 1.6.7, 1.10.5, and 1.11.3

maltezfaria:
Thanks for the fixes, the tests now pass!

I am trying to understand something rather basic here (have not played much with precompilation). IIUC, `__precompile__(true)` is the default, and at present when I do `using DataFlowTasks` a first time it does trigger precompilation (which throws no errors). So it is not clear to me why these changes are needed. In your original post you said that `DataFlowTasks` uses `precompile(false)`, but is that something we currently do implicitly?

camilodlt:
I think that it turns out that I was using DataFlowTasks 0.1.0 in which there is a method overwrite, which prevents precompilation and I thought that a `__precompile__()` call would solve it. In that case, my problem relates to the dependencies which prevent the project to update to the 0.2.0 which is precompilable as is. In that case you can disregard my PR 😅

maltezfaria:
> I think that it turns out that I was using DataFlowTasks 0.1.0 in which there is a method overwrite, which prevents precompilation and I thought that a `__precompile__()` call would solve it. In that case, my problem relates to the dependencies which prevent the project to update to the 0.2.0 which is precompilable as is. In that case you can disregard my PR 😅

What dependencies are preventing the upgrade? Maybe DataFlowTasks is actually compatible with older versions of the packages listed in the `Project.toml`, in which case we could loosen the `compat` entry and tag a minor release so that you can update.

camilodlt:
I was able to update to 0.2.0 in my project by calling `Pkg.add(Pkg.PackageSpec(;name="DataFlowTasks", version="0.2.0"))` directly. I don't know why a simple `] update` was not enough to update the package. I think that in that case I have to add a compat on my end requiring at least 0.2.0.

maltezfaria:
> I was able to update to 0.2.0 in my project by calling `Pkg.add(Pkg.PackageSpec(;name="DataFlowTasks", version="0.2.0"))` directly. I don't know why a simple `] update` was not enough to update the package. I think that in that case I have to add a compat on my end requiring at least 0.2.0.

If for some reason you had a `compat` entry of DataFlowTasks pinned to `0.1` it won't automatically update to `0.2` because it is considered a breaking change. I recommend using [CompatHelper](https://github.com/JuliaRegistries/CompatHelper.jl) to keep track of that! 

Should we close this?

camilodlt:
I think the problem is that in my project I had Makie v0.22.1 which is not accepted in your compat which is checked even if I don't use the extras. Adding the specific version works by downgrading Makie, GLMakie in my case.

